PR #1, opened 2021-11-03
Fix sillsdev/silnlp#96 - make compatible with S3 bucket.

johnml1135:


machine/corpora/paratext_text_corpus.py
```diff
@@ -22,7 +22,8 @@ def __init__(
         if not settings_filename.is_file():
             raise RuntimeError("The project directory does not contain a settings file.")
 
-        settings_tree = etree.parse(str(settings_filename))
+        with settings_filename.open("rb") as stream:
```
ddaspit, line 25: How does this change make it compatible with S3 buckets?

johnml1135: instead of handing the path, it opens it using S3Path.  It's the same change that works with silnlp.

ddaspit: The constructor can accept a `str` or a `Path` object for the `project_dir` parameter. To ensure that `project_dir` is a `Path` object, it constructs a `Path` object. If an `S3Path` is passed in, this will turn it into a normal `Path` object. You probably need to add in some code that only constructs a new `Path` object if `project_dir` is a `str`.

ddaspit (changes requested):

ddaspit:
I went ahead and implemented this. This PR is no longer needed.

PR #29, opened 2023-09-14
catch all errors and mark clearml task as failed with status_reason

mshannon-sil:
I added a try except block that will first log any error and then, if it's running on clearml, will call `mark_failed()` with the error message as the `status_reason`. This `status_reason` can then be accessed by machine to pass along to the user.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)

### DIFF
--- a/machine/jobs/build_nmt_engine.py
+++ b/machine/jobs/build_nmt_engine.py
@@ -58,9 +58,9 @@ def run(args: dict) -> None:
         job.run(check_canceled)
         logger.info("Finished")
     except Exception as e:
-        logger.exception(str(e))
+        logger.exception(e, stack_info=True)
         if task:
-            task.mark_failed(status_reason=str(e))
+            task.mark_failed(status_reason=type(e).__name__, status_message=str(e))
 
 
 def main() -> None:
--- a/machine/jobs/build_nmt_engine.py
+++ b/machine/jobs/build_nmt_engine.py
@@ -22,6 +22,7 @@ logger = logging.getLogger(__package__ + ".build_nmt_engine")
 
 def run(args: dict) -> None:
     check_canceled: Optional[Callable[[], None]] = None
+    task = None
     if args["clearml"]:
         task = Task.init()
 
@@ -31,30 +32,35 @@ def run(args: dict) -> None:
 
         check_canceled = clearml_check_canceled
 
-    logger.info("NMT Engine Build Job started")
+    try:
+        logger.info("NMT Engine Build Job started")
 
-    SETTINGS.update(args)
-    SETTINGS.data_dir = os.path.expanduser(cast(str, SETTINGS.data_dir))
+        SETTINGS.update(args)
+        SETTINGS.data_dir = os.path.expanduser(cast(str, SETTINGS.data_dir))
 
-    logger.info(f"Config: {SETTINGS.as_dict()}")
+        logger.info(f"Config: {SETTINGS.as_dict()}")
 
-    shared_file_service = ClearMLSharedFileService(SETTINGS)
-    model_type = cast(str, SETTINGS.model_type).lower()
-    nmt_model_factory: NmtModelFactory
-    if model_type == "huggingface":
-        from .huggingface.hugging_face_nmt_model_factory import HuggingFaceNmtModelFactory
+        shared_file_service = ClearMLSharedFileService(SETTINGS)
+        model_type = cast(str, SETTINGS.model_type).lower()
+        nmt_model_factory: NmtModelFactory
+        if model_type == "huggingface":
+            from .huggingface.hugging_face_nmt_model_factory import HuggingFaceNmtModelFactory
 
-        nmt_model_factory = HuggingFaceNmtModelFactory(SETTINGS, shared_file_service)
-    elif model_type == "opennmt":
-        from .opennmt.open_nmt_model_factory import OpenNmtModelFactory
+            nmt_model_factory = HuggingFaceNmtModelFactory(SETTINGS, shared_file_service)
+        elif model_type == "opennmt":
+            from .opennmt.open_nmt_model_factory import OpenNmtModelFactory
 
-        nmt_model_factory = OpenNmtModelFactory(SETTINGS, shared_file_service)
-    else:
-        raise RuntimeError("The model type is invalid.")
+            nmt_model_factory = OpenNmtModelFactory(SETTINGS, shared_file_service)
+        else:
+            raise RuntimeError("The model type is invalid.")
 
-    job = NmtEngineBuildJob(SETTINGS, nmt_model_factory, shared_file_service)
-    job.run(check_canceled)
-    logger.info("Finished")
+        job = NmtEngineBuildJob(SETTINGS, nmt_model_factory, shared_file_service)
+        job.run(check_canceled)
+        logger.info("Finished")
+    except Exception as e:
+        logger.exception(str(e))
+        if task:
+            task.mark_failed(status_reason=str(e))
 
 
 def main() -> None:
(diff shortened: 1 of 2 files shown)


mshannon-sil:
I see that the checks are failing. I'll remove the review requests until they're passing.

johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 63 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIQSFJE2bJPyU7olLK:-NeIQSFJE2bJPyU7olLL:b-mpbb1h) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L63)):*
> ```Python
>         logger.exception(str(e))
>         if task:
>             task.mark_failed(status_reason=str(e))
> ```

Wouldn't it be better for `status_reason=type(exception).__name__,status_message=str(e)`

johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIRQOR4sMLRLw0YRGb:-NeIRQOR4sMLRLw0YRGc:b-qb5zzo) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L61)):*
> ```Python
>         logger.info("Finished")
>     except Exception as e:
>         logger.exception(str(e))
> ```

Should this be if task, then mark_failed, else logger?  If tasks exists, does it double-log the error?

johnml1135 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @mshannon-sil)

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/jobs/build_nmt_engine.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIRQOR4sMLRLw0YRGb:-NeIrcdv0DN17jOBOZ7Y:b-ajqb5y) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L61)):*

Should this be if task, then mark_failed, else logger?  If tasks exists, does it double-log the error?
</blockquote></details>

`mark_failed()` does not log any output to the console, it just sets the clearml task as failed and sets the corresponding `status_reason` and `status_message` of the task. That's why I had the logger run in all cases. However, something I was considering was simply raising the exception again at the end of the except block rather than logging the error, since raising the exception prints the full traceback unlike the logger which only includes the final line in the traceback. Would this be preferable in your opinion?
___
*[`machine/jobs/build_nmt_engine.py` line 63 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIQSFJE2bJPyU7olLK:-NeIqs-BB_lbnEONfn2y:bmsxghq) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L63)):*

Wouldn't it be better for `status_reason=type(exception).__name__,status_message=str(e)`
</blockquote></details>

Good idea, I'll make that change.

johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIRQOR4sMLRLw0YRGb:-NeKI5hi9qQTYlTSGvAN:bkpo9jk) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L61)):*

`mark_failed()` does not log any output to the console, it just sets the clearml task as failed and sets the corresponding `status_reason` and `status_message` of the task. That's why I had the logger run in all cases. However, something I was considering was simply raising the exception again at the end of the except block rather than logging the error, since raising the exception prints the full traceback unlike the logger which only includes the final line in the traceback. Would this be preferable in your opinion?
</blockquote></details>

I would prefer the whole traceback in the log - it is for diagnosis above all.  The whole traceback is not needed in mark_failed - as that is more for the user than the engineer.

johnml1135 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @mshannon-sil)

ddaspit (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`machine/jobs/build_nmt_engine.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIRQOR4sMLRLw0YRGb:-NeKj8Lq1JfqgAXa5Fql:b-4vn1nh) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L61)):*

I would prefer the whole traceback in the log - it is for diagnosis above all.  The whole traceback is not needed in mark_failed - as that is more for the user than the engineer.
</blockquote></details>

I was under the impression that `logger.exeception` prints the full stack trace.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`machine/jobs/build_nmt_engine.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine.py/29#-NeIRQOR4sMLRLw0YRGb:-NeKqpMh2vmvHlK3s1GP:b18ny) ([raw file](https://github.com/sillsdev/machine.py/blob/98ce526e8c3c726645ec18cdf55aaec2863ad68d/machine/jobs/build_nmt_engine.py#L61)):*

I was under the impression that `logger.exeception` prints the full stack trace.
</blockquote></details>

The default behavior for `logger.exception()` only prints out the final line in the stack trace, but I looked into it and saw that if you set `stack_trace=True` it will print the whole stack trace. I went ahead and added that in my most recent commit. Also, I discovered that if I call `raise` at the end of the except block, it somehow is overwriting the `status_reason` and `status_message` that I set for the task, so using raise wasn't a tenable solution anyways.

mshannon-sil:
@johnml1135 @ddaspit All of the review comments have been addressed now. And it looks like the checks failing isn't related to any of the code I added, since they're failing for the current commit in main too. I can do a separate pull request investigating those issues, but for now I think we should go forward with merging the error reporting into main.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29#-:-NeP0E2Q6IAxStY1H_LN:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

johnml1135 (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/29)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #31, opened 2023-09-22
Update regex

ddaspit:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/31)


johnml1135:
Why did regex need to be updated?

ddaspit:
We needed a newer version that had Python 3.11 wheels.

PR #43, opened 2023-10-13
Tokenizer

mshannon-sil:
Added some code and tests for adding missing characters into machine.py.

Along the way, I ran into two issues related to existing code for adding language codes. The first issue was that if both lang_codes were new, it would assign the wrong index for the second one and then crash. I fixed this by upgrading to the latest version of transformers. However, the other issue is that for every tokenizer I've tried that isn't NllbTokenizerFast, added language codes don't save when the tokenizer is saved, and so when the tokenizer is reloaded it crashes when given the new language code. I haven't been able to resolve that yet, so I'm going to open a separate issue for it.

Also, in my tests I'm using the model `hf-internal-testing/tiny-random-nllb`. However, I think there's a mistake in the [tokenizer_config.json](https://huggingface.co/hf-internal-testing/tiny-random-nllb/blob/main/tokenizer_config.json)) file, because it has `"tokenizer_class": "MBartTokenizer"` when it should be `"tokenizer_class": "NllbTokenizer"`. Should I go ahead and submit a pull request on huggingface to change that, or is there a reason I'm not aware of why it should be MBartTokenizer?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)


johnml1135:
*[`machine/jobs/settings.yaml` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsR7MO0sJcvPamBlPI:-NgsR7MO0sJcvPamBlPJ:b-2wkwot) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/jobs/settings.yaml#L8)):*
> ```YAML
>   tokenizer:
>     update_src: false
>     update_trg: false
> ```

Why are we defaulting this to false?  Wouldn't the default behavior be to update the tokenizer?  Is there some reason we would want to default to false?

johnml1135:
*[`machine/tokenization/custom_normalizer/tokenizer_config.json` line 2 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsSKLS-G0wPtthblWO:-NgsSKLS-G0wPtthblWP:b-dybscp) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/tokenization/custom_normalizer/tokenizer_config.json#L2)):*
> ```JSON
> {
>   "additional_special_tokens": null,
> ```

Can this be modified through the build_params?

johnml1135:
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 175 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsSk9J-5yW8uFZSJiH:-NgsSk9J-5yW8uFZSJiI:b-i02yny) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L175)):*
> ```Python
>                     data["model"]["vocab"][token] = vocab_len + i
>                 file.seek(0)
>                 json.dump(data, file, ensure_ascii=False, indent=4)
> ```

Why are you overwriting the tokenizer.json file?  Why not create a new file in the project (or potentially, a small json diff that would merged into tokenizer.json...)?

johnml1135 (commented):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @mshannon-sil)

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/jobs/settings.yaml` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsR7MO0sJcvPamBlPI:-Ngu9beA-G_PDH_4-YoD:b-m9vh9g) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/jobs/settings.yaml#L8)):*

Why are we defaulting this to false?  Wouldn't the default behavior be to update the tokenizer?  Is there some reason we would want to default to false?
</blockquote></details>

Good catch, I just defaulted them to `false` since that's essentially what SILNLP does, but now that we've tested missing characters extensively we want `true` to be the default.
___
*[`machine/tokenization/custom_normalizer/tokenizer_config.json` line 2 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsSKLS-G0wPtthblWO:-Ngu9zrMB6XVRUD8UeLT:b-4qlvud) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/tokenization/custom_normalizer/tokenizer_config.json#L2)):*

Can this be modified through the build_params?
</blockquote></details>

No need, this file is only needed to overwrite for the normalizer it contains, no other part of it is used.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 175 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsSk9J-5yW8uFZSJiH:-NguADgq5KbrQk2jij_O:b-l3aq0q) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L175)):*

Why are you overwriting the tokenizer.json file?  Why not create a new file in the project (or potentially, a small json diff that would merged into tokenizer.json...)?
</blockquote></details>

Well there's no need to keep the older tokenizer file, so I think it would be a bit of a waste of memory space to have an old and new version of tokenizer.json in the same project directory. As for merging a small json diff, I didn't think there was a way to get around overwriting the whole file. All the recommendations I see online just dump the whole updated json object into the file, overwriting all its contents. Do you have a link to an example of merging json into a json file in place without overwriting everything?

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: 5 of 8 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/jobs/settings.yaml` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsR7MO0sJcvPamBlPI:-Ngv-hWI2RvkqgxL7EIL:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/jobs/settings.yaml#L8)):*

Good catch, I just defaulted them to `false` since that's essentially what SILNLP does, but now that we've tested missing characters extensively we want `true` to be the default.
</blockquote></details>

Done.

johnml1135:
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 175 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsSk9J-5yW8uFZSJiH:-NgxTeI4-5-iLiVzQ4Oc:b-fdx6u3) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L175)):*

Well there's no need to keep the older tokenizer file, so I think it would be a bit of a waste of memory space to have an old and new version of tokenizer.json in the same project directory. As for merging a small json diff, I didn't think there was a way to get around overwriting the whole file. All the recommendations I see online just dump the whole updated json object into the file, overwriting all its contents. Do you have a link to an example of merging json into a json file in place without overwriting everything?
</blockquote></details>

That's fine - as long as it worked.  The biggest concern would be if there were two successive runs using machine.py - the tokenizer.json would continue to drift over time.  I am assuming that would not be desired, but neither would it be a huge issue either.  Also, we run these jobs on a fresh docker image each time.  It would have more concern for incremental tuning - if we need to add more and more tokens over time.

johnml1135 (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @mshannon-sil)

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: 3 of 8 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @johnml1135)

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 175 at r1](https://reviewable.io/reviews/sillsdev/machine.py/43#-NgsSk9J-5yW8uFZSJiH:-NgyBwhp6y4tYkAMXeTH:b9dtg3g) ([raw file](https://github.com/sillsdev/machine.py/blob/c4bb2b8030a12b7e172e43598a1101cf8bfda8bf/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L175)):*

That's fine - as long as it worked.  The biggest concern would be if there were two successive runs using machine.py - the tokenizer.json would continue to drift over time.  I am assuming that would not be desired, but neither would it be a huge issue either.  Also, we run these jobs on a fresh docker image each time.  It would have more concern for incremental tuning - if we need to add more and more tokens over time.
</blockquote></details>

Each time we run a new job, it starts with an unmodified tokenizer, not from the tokenizer files saved in the directory, so we shouldn't run into any issues with drift.

ddaspit (changes requested):
> Reviewed 2 of 8 files at r1, 1 of 3 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 5 unresolved discussions (waiting on @mshannon-sil)

___
*[`machine/tokenization/custom_normalizer/tokenizer_config.json` line 1 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2ZoPw349eUKTAAXhx:-Nh2ZoPw349eUKTAAXhy:b3g1noh) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/tokenization/custom_normalizer/tokenizer_config.json#L1)):*
> ```JSON
> {
> ```

I would move this file over to the huggingface directory.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 87 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2YVtA4IkCU7w97esg:-Nh2YVtA4IkCU7w97esh:b-huo7bz) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L87)):*
> ```Python
>         max_source_length: Optional[int] = None,
>         max_target_length: Optional[int] = None,
>         config: Union[LazySettings, dict] = {},
> ```

Rather than using a generic config dictionary, just add explicit parameters for turning on adding missing characters. It should be turned on by default for target and turned off by default for source.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 161 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2ZNNN0RmMMe5yvBoT:-Nh2ZNNN0RmMMe5yvBoU:by3a8oc) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L161)):*
> ```Python
>             mpn.substitutions = [(re.compile(r), sub) for r, sub in mpn.substitutions]
>             charset = {mpn.normalize(char) for char in charset}
>             charset = {tokenizer.backend_tokenizer.normalizer.normalize_str(char) for char in charset}  # type: ignore
> ```

Why are you ignoring the type error here? When we run into type errors, we want to fix them properly if possible.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 193 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_beW4bRXwsYpfJQs:-Nh2_beW4bRXwsYpfJQt:b8awgln) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L193)):*
> ```Python
>                 )
>             else:
>                 norm_tok = PreTrainedTokenizerFast.from_pretrained(
> ```

What is `norm_tok` used for?
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-Nh2_hgo74b2NEZIVbNw:b-a54qi6) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*
> ```Python
>                     "./machine/tokenization/custom_normalizer", use_fast=True
>                 )
>                 tokenizer.backend_tokenizer.normalizer = norm_tok.backend_tokenizer.normalizer  # type: ignore
> ```

What are the type errors here? Can they be fixed?

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: 2 of 8 files reviewed, 5 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 87 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2YVtA4IkCU7w97esg:-Nh3Z0LN9MsJF4zwxS8W:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L87)):*

Rather than using a generic config dictionary, just add explicit parameters for turning on adding missing characters. It should be turned on by default for target and turned off by default for source.
</blockquote></details>

Done.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 161 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2ZNNN0RmMMe5yvBoT:-Nh3ZgxaDOhe8DgMPd47:b-se9yun) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L161)):*

Why are you ignoring the type error here? When we run into type errors, we want to fix them properly if possible.
</blockquote></details>

Well I remember investigating it and not figuring out what could be causing the type error and that there really shouldn't be a type error, so I added that ignore flag. However I just removed the flag, and there isn't a type error anymore, so some other code I wrote must have addressed it. Point noted about fixing them properly, I'll avoid using `# type: ignore` unless absolutely necessary.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 193 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_beW4bRXwsYpfJQs:-Nh3Z4ar-mn6IpX0HHCg:b-xd0z7w) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L193)):*

What is `norm_tok` used for?
</blockquote></details>

It's just used to give the actual tokenizer the custom normalizer that doesn't normalize away ZWJ, ZWNJ, and Glottal Stop characters,
___
*[`machine/tokenization/custom_normalizer/tokenizer_config.json` line 1 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2ZoPw349eUKTAAXhx:-Nh30GlUBc792OChyPQh:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/tokenization/custom_normalizer/tokenizer_config.json#L1)):*

I would move this file over to the huggingface directory.
</blockquote></details>

Done.

ddaspit (commented):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @mshannon-sil)

___
*[`machine/jobs/settings.yaml` line 6 at r4](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh3t4h326iytcwsaeHH:-Nh3t4h326iytcwsaeHI:bgq6ohu) ([raw file](https://github.com/sillsdev/machine.py/blob/d3738f316fdbafde8c125f96dfc46890a8c50a5d/machine/jobs/settings.yaml#L6)):*
> ```YAML
>   data_dir: ~/machine
>   batch_size: 1024
>   tokenizer:
> ```

This should be moved under `huggingface`.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 86 at r4](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh3rOWf7i1kSxBx2ZuF:-Nh3rOWf7i1kSxBx2ZuG:b115fl) ([raw file](https://github.com/sillsdev/machine.py/blob/d3738f316fdbafde8c125f96dfc46890a8c50a5d/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L86)):*
> ```Python
>         max_source_length: Optional[int] = None,
>         max_target_length: Optional[int] = None,
>         update_src: bool = False,
> ```

I think a more descriptive name would be better. Maybe `add_unk_src_tokens`.
___
*[`machine/tokenization/custom_normalizer/tokenizer_config.json` line 1 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2ZoPw349eUKTAAXhx:-Nh3sWNkEWJFNT2gQMqr:bz35289) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/tokenization/custom_normalizer/tokenizer_config.json#L1)):*

Done.
</blockquote></details>

I would put this in a subdirectory of `huggingface` similar to what you did before.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-Nh3pAOZ9woV9VNdN3Ec:b-e58ikb) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*

What are the type errors here? Can they be fixed?
</blockquote></details>

The first error is `Property "normalizer" has no defined setter`. However, some huggingface documentation https://github.com/huggingface/tokenizers/blob/b24a2fc1781d5da4e6ebcd3ecb5b91edffc0a05f/bindings/python/examples/custom_components.py uses this method of overwriting the normalizer, and I've tested that it works correctly, so there shouldn't be an issue. The second and third errors say `Cannot access member "source_corpus" for type "Dataset"
  Member "source_corpus" is unknown` and `Cannot access member "target_corpus" for type "Dataset"
  Member "target_corpus" is unknown` respectively. Not sure why this just started appearing now since I didn't write that code and I don't think I changed anything that should have affected it. Any recommendations for how to resolve these?

mshannon-sil:
*[`machine/jobs/settings.yaml` line 6 at r4](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh3t4h326iytcwsaeHH:-Nh3wp97EZ0J3Tz_jVdT:b-aui1i7) ([raw file](https://github.com/sillsdev/machine.py/blob/d3738f316fdbafde8c125f96dfc46890a8c50a5d/machine/jobs/settings.yaml#L6)):*

This should be moved under `huggingface`.
</blockquote></details>

Should I also change these names to `add_unk_src_tokens` and `add_unk_trg_tokens` to match the changed argument names in HuggingfaceNmtModelTrainer?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @mshannon-sil)

___
*[`machine/jobs/settings.yaml` line 6 at r4](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh3t4h326iytcwsaeHH:-Nh3zUxB2Dft5yp5DkXQ:bw6ors) ([raw file](https://github.com/sillsdev/machine.py/blob/d3738f316fdbafde8c125f96dfc46890a8c50a5d/machine/jobs/settings.yaml#L6)):*

Should I also change these names to `add_unk_src_tokens` and `add_unk_trg_tokens` to match the changed argument names in HuggingfaceNmtModelTrainer?
</blockquote></details>

Yes, it would make sense to keep them consistent.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-Nh3zhXa31jh0Q2cijYw:b-pmgqz2) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*

The first error is `Property "normalizer" has no defined setter`. However, some huggingface documentation https://github.com/huggingface/tokenizers/blob/b24a2fc1781d5da4e6ebcd3ecb5b91edffc0a05f/bindings/python/examples/custom_components.py uses this method of overwriting the normalizer, and I've tested that it works correctly, so there shouldn't be an issue. The second and third errors say `Cannot access member "source_corpus" for type "Dataset"
  Member "source_corpus" is unknown` and `Cannot access member "target_corpus" for type "Dataset"
  Member "target_corpus" is unknown` respectively. Not sure why this just started appearing now since I didn't write that code and I don't think I changed anything that should have affected it. Any recommendations for how to resolve these?
</blockquote></details>

I will take a look at them and see if I can figure out where the errors are coming from.

johnml1135:
*[`machine/jobs/settings.yaml` line 24 at r5](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh7PjX19hLBmVDLhqeC:-Nh7PjX19hLBmVDLhqeD:b5c9ec3) ([raw file](https://github.com/sillsdev/machine.py/blob/ce9b12f223be9cbdf95641bd8538e99502128930/machine/jobs/settings.yaml#L24)):*
> ```YAML
>       batch_size: 16
>     tokenizer:
>       add_unk_src_tokens: false
> ```

Why false and true?  What if people are doing a Backtranslation?  Why would we not want both to be true?

johnml1135:
*[`tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py` line 203 at r5](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh7Q8J8EXY9dsjtO_jZ:-Nh7Q8J95wKfIjZmMGl0:b77u8xe) ([raw file](https://github.com/sillsdev/machine.py/blob/ce9b12f223be9cbdf95641bd8538e99502128930/tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py#L203)):*
> ```Python
>             max_target_length=20,
>             add_unk_src_tokens=False,
>             add_unk_trg_tokens=False,
> ```

Sometimes initialized to true sometimes to false - why?

johnml1135 (commented):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 5 unresolved discussions (waiting on @mshannon-sil)

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/jobs/settings.yaml` line 6 at r4](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh3t4h326iytcwsaeHH:-Nh4BPUG7VSf73Hk-6Mo:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/d3738f316fdbafde8c125f96dfc46890a8c50a5d/machine/jobs/settings.yaml#L6)):*

Yes, it would make sense to keep them consistent.
</blockquote></details>

Done.
___
*[`machine/jobs/settings.yaml` line 24 at r5](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh7PjX19hLBmVDLhqeC:-Nh7Wemp1RQENA07HZih:bcva1yq) ([raw file](https://github.com/sillsdev/machine.py/blob/ce9b12f223be9cbdf95641bd8538e99502128930/machine/jobs/settings.yaml#L24)):*

Yeah, we should probably turn them both on by default for Serval.
</blockquote></details>

I agree that we should always want to add missing characters whether they're in the source or the target. I'll set them back to being both true.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-Nh7WQnpD_kEjoMoM79P:blr7omc) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*

Did you fix these type issues?
</blockquote></details>

No not yet, the last commit was to address the other comments.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 86 at r4](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh3rOWf7i1kSxBx2ZuF:-Nh4BQ_D9fyvLCnwmKPb:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/d3738f316fdbafde8c125f96dfc46890a8c50a5d/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L86)):*

I think a more descriptive name would be better. Maybe `add_unk_src_tokens`.
</blockquote></details>

Done.
___
*[`tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py` line 203 at r5](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh7Q8J8EXY9dsjtO_jZ:-Nh7WKtQ6mTuHNdnpozN:b8nxv5e) ([raw file](https://github.com/sillsdev/machine.py/blob/ce9b12f223be9cbdf95641bd8538e99502128930/tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py#L203)):*

Sometimes initialized to true sometimes to false - why?
</blockquote></details>

This is in the test directory, so I have different settings to make sure it works as expected.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-Nh8Y20y2drlaqI7lf04:b9q71n8) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*

No not yet, the last commit was to address the other comments.
</blockquote></details>

For the normalizer type error, I think we should just leave it like that. Any changes would be on Huggingface's end to implement a setter function, but it works regardless. In addition to the huggingface example I linked to earlier demonstrating a custom Normalizer replacing the original, there is this stack overflow answer where someone else did the same thing (https://stackoverflow.com/questions/74490128/nlp-huggingface-tokenizer-wont-let-me-set-pre-tokenizer-property).

As for the two corpus type errors, what's going on is that the type of self._corpus is defined as either Dataset or ParallelTextCorpus. I realized that my code is designed to work for ParallelTextCorpus and would fail if it receives a Dataset type corpus. However, looking at the codebase, I can't find a single instance where Dataset would be passed in instead of ParallelTextCorpus, and in my tests they have all been ParallelTextCorpus. I think it would be simpler to just set self._corpus to ParallelTextCorpus exclusively, unless there's a need for the Dataset type to be passed in.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: all files reviewed, 4 unresolved discussions (waiting on @johnml1135 and @mshannon-sil)

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-Nh8jfmv2UosxXYWagl-:b1m82h1) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*

For the normalizer type error, I think we should just leave it like that. Any changes would be on Huggingface's end to implement a setter function, but it works regardless. In addition to the huggingface example I linked to earlier demonstrating a custom Normalizer replacing the original, there is this stack overflow answer where someone else did the same thing (https://stackoverflow.com/questions/74490128/nlp-huggingface-tokenizer-wont-let-me-set-pre-tokenizer-property).

As for the two corpus type errors, what's going on is that the type of self._corpus is defined as either Dataset or ParallelTextCorpus. I realized that my code is designed to work for ParallelTextCorpus and would fail if it receives a Dataset type corpus. However, looking at the codebase, I can't find a single instance where Dataset would be passed in instead of ParallelTextCorpus, and in my tests they have all been ParallelTextCorpus. I think it would be simpler to just set self._corpus to ParallelTextCorpus exclusively, unless there's a need for the Dataset type to be passed in.
</blockquote></details>

Regarding the normalizer type error, I think ignoring the error is fine. We should add a comment explaining that we are using undocumented/unsupported behavior.

Although none of our code uses `Dataset`, you have to remember that Machine.py is a shared library. I would prefer not to remove this functionality, since it is useful to other applications that might embed Machine.py. I don't think it would be too difficult to update the code to support `Dataset` as well. You can look at the `preprocess_function` for an example of how to extract the source and target segments from the `Dataset`.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 157 at r5](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh8pBll9hJ164QyPMYI:-Nh8pBlmC3pXOiHfX0Hm:b-gk8x5t) ([raw file](https://github.com/sillsdev/machine.py/blob/ce9b12f223be9cbdf95641bd8538e99502128930/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L157)):*
> ```Python
>             charset = set()
>             for text in texts:
>                 for row in text._get_rows():
> ```

You should use the `get_rows` method instead of the internal `_get_rows` method. The code would look like this:
```
for text in texts:
    with text.get_rows() as rows:
        for row in rows:
            charset = charset | set(row.text)
```
Although you don't have to use the context manager (`with` statement), it is safer to use it, since it will ensure that any underlying files/streams are closed.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43)* status: 5 of 9 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 196 at r3](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh2_hgo74b2NEZIVbNv:-NhCu8Az9t9gdoiCAvaa:b-6i0rsf) ([raw file](https://github.com/sillsdev/machine.py/blob/e3dc3635f074b964ef1e9e9ae5a1b382a2b68274/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L196)):*

Regarding the normalizer type error, I think ignoring the error is fine. We should add a comment explaining that we are using undocumented/unsupported behavior.

Although none of our code uses `Dataset`, you have to remember that Machine.py is a shared library. I would prefer not to remove this functionality, since it is useful to other applications that might embed Machine.py. I don't think it would be too difficult to update the code to support `Dataset` as well. You can look at the `preprocess_function` for an example of how to extract the source and target segments from the `Dataset`.
</blockquote></details>

Okay I added an explanatory comment.

I ended up moving the creation of the `train_dataset` variable from self._corpus to an earlier point in the code, so that I can just pass `train_dataset` into `missing_characters()` along with the language codes corresponding to the src/trg needing updating. I correspondingly refactored `missing_characters()` to support the Dataset type instead of the List[Text] type. I figured this was more streamlined than having the `missing_characters()` function need to support both Datasets and List[Text], and I don't think that moving the creation of the `train_dataset` has any negative impacts on the code. But if stylistically you'd prefer to keep the creation of `train_dataset` where it was, I can refactor the function again to support both types.
___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 157 at r5](https://reviewable.io/reviews/sillsdev/machine.py/43#-Nh8pBll9hJ164QyPMYI:-NhCyZNAD4D6w3hS2xed:b-ljfvff) ([raw file](https://github.com/sillsdev/machine.py/blob/ce9b12f223be9cbdf95641bd8538e99502128930/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L157)):*

You should use the `get_rows` method instead of the internal `_get_rows` method. The code would look like this:
```
for text in texts:
    with text.get_rows() as rows:
        for row in rows:
            charset = charset | set(row.text)
```
Although you don't have to use the context manager (`with` statement), it is safer to use it, since it will ensure that any underlying files/streams are closed.
</blockquote></details>

Changed to work with Dataset rather than List[Text].

ddaspit (approved):
:lgtm:

> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43#-:-NhD0i2nCUipLb2TSpY0:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

johnml1135 (approved):
:lgtm:

> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/43#-:-NhDEJ2S-R19oqGFannY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #57, opened 2023-11-09
handle stopped tasks without throwing error

mshannon-sil:
This PR ensures that we catch exceptions caused by stopping a clearml task without reraising the exception or marking the task as failed. Also, the `check_canceled()` function is now called before progress updates are made in the `_ProgressCallback` methods rather than after.

In addition, I removed the `stopping` status in the `check_canceled()` function since it's not listed under `task_status_enum`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/57)

### DIFF
--- a/machine/jobs/build_nmt_engine.py
+++ b/machine/jobs/build_nmt_engine.py
@@ -30,7 +30,7 @@ def run(args: dict) -> None:
         task = Task.init()
 
         def clearml_check_canceled() -> None:
-            if task.get_status() in {"stopped", "stopping"}:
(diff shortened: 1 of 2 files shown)


johnml1135 (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/57)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #61, opened 2023-11-10
Move get_books to parse.py and add get_chapters

isaac091:
Created new module for get_books and get_chapters due to circular import conflict.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)


ddaspit (changes requested):
> Reviewed 3 of 5 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)* status: all files reviewed, 7 unresolved discussions (waiting on @isaac091 and @johnml1135)

___
*[`machine/scripture/parse.py` line 44 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu3j7B9djEBjd5tCXs:-Niu3j7B9djEBjd5tCXt:b-o9kbkn) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L44)):*
> ```Python
> # Output format: { book_num: [chapters] }
> # An empty list, i.e. book_num: [] signifies the inclusion of all chapters, while the absence of an entry means the book is not included
> def get_chapters(chapter_selections: str) -> dict:
> ```

The return type should be more explicit.
___
*[`machine/scripture/parse.py` line 45 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu4Dya7DT_2u7cBVjX:-Niu4Dya7DT_2u7cBVjY:b-dgrjig) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L45)):*
> ```Python
> # An empty list, i.e. book_num: [] signifies the inclusion of all chapters, while the absence of an entry means the book is not included
> def get_chapters(chapter_selections: str) -> dict:
>     versification = Versification.create("Original")
> ```

You should pass in the versification as an argument.
___
*[`machine/scripture/parse.py` line 51 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu3DT0BrKfCTMoqjY_:-Niu3DT0BrKfCTMoqjYa:b-wt2f2x) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L51)):*
> ```Python
> 
>     # Normalize books written as "MAT01.usfm" to "MAT"
>     chapter_selections = re.sub(USFM_FILE_PATTERN, "", chapter_selections)
> ```

This seems specific to silnlp. You should do this in silnlp.
___
*[`machine/scripture/parse.py` line 72 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu7qcrEe-USpEZSV7d:-Niu7qcrEe-USpEZSV7e:b-poax37) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L72)):*
> ```Python
> 
>             if book == 0:
>                 raise RuntimeError(f"{section[:3]} is an invalid book ID.")
> ```

I would use `ValueError` for all exceptions.
___
*[`machine/scripture/parse.py` line 77 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu7vuB-1Ili_mztYMZ:-Niu7vuB-1Ili_mztYM_:b-90y0so) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L77)):*
> ```Python
>             chapters[book] = set()
>             last_chapter = versification.get_last_chapter(book)
>             for chapter_num in chapter_nums:
> ```

We should raise an exception if the chapter is not valid, i.e. out of range.
___
*[`machine/scripture/parse.py` line 87 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu5VYc6eqvt5pGteDl:-Niu5VYc6eqvt5pGteDm:bj7brpy) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L87)):*
> ```Python
>             # Delete entry if no chapter numbers were valid
>             if len(chapters[book]) == 0:
>                 del chapters[book]
> ```

It would probably be clearer to only add the set if it is not empty.
___
*[`machine/scripture/parse.py` line 89 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu6NXt7t8HLA_NT-Zw:-Niu6NXt7t8HLA_NT-Zx:b-qphrax) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L89)):*
> ```Python
>                 del chapters[book]
>         elif "-" in section:  # Spans of books
>             spans.append(section)
> ```

Is there some reason why you don't just process the span here?

isaac091:
I don't process the spans right away to prioritize getting specific chapters from a book, so for example if someone wanted "MAT-ROM;ACT1-3", they wouldn't have to worry about putting the Acts chapters first to avoid getting the selections overwritten with the span. But I guess that's not consistent all the way through because if "NT" or "OT" are passed first, all of those books are added right away. I can change it so that there's no "order of operations", or if we want to keep it this way, I'll change it so that "NT" and "OT" are added last as well.

johnml1135:
The order shouldn't matter - I would keep it as original versification order.  Are you accounting for removals as well?

johnml1135:
*[`machine/scripture/parse.py` line 8 at r3](https://reviewable.io/reviews/sillsdev/machine.py/61#-NiuGVDu2e7fwBamliCu:-NiuGVDu2e7fwBamliCv:bv5y36a) ([raw file](https://github.com/sillsdev/machine.py/blob/55c982dd528aba15e556a1023a8a171443b2318e/machine/scripture/parse.py#L8)):*
> ```Python
> 
> USFM_FILE_PATTERN = re.compile(r"(?<=[A-Z]{3})\d+\.usfm")
> BOOK_SPAN = re.compile(r"[A-Z]{3}-[A-Z]{3}")
> ```

Books can include numbers - such as 1CO.

johnml1135:
*[`machine/scripture/parse.py` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine.py/61#-NiuGZnKEFmWmln9A3I5:-NiuGZnKEFmWmln9A3I6:b-luigjt) ([raw file](https://github.com/sillsdev/machine.py/blob/55c982dd528aba15e556a1023a8a171443b2318e/machine/scripture/parse.py#L7)):*
> ```Python
> from .verse_ref import Versification
> 
> USFM_FILE_PATTERN = re.compile(r"(?<=[A-Z]{3})\d+\.usfm")
> ```

Books include numbers.

johnml1135:
*[`machine/scripture/parse.py` line 13 at r3](https://reviewable.io/reviews/sillsdev/machine.py/61#-NiuGjUo6OpJqUx-KOC8:-NiuGjUo6OpJqUx-KOC9:b-ptv8nb) ([raw file](https://github.com/sillsdev/machine.py/blob/55c982dd528aba15e556a1023a8a171443b2318e/machine/scripture/parse.py#L13)):*
> ```Python
> def get_books(books: Union[str, List[str]]) -> Set[int]:
>     if isinstance(books, str):
>         books = books.split(",")
> ```

Book chapter lists can have multiple parts, such as MAT1-3, 5, 7, 10-12.  Because of this, we need to use a ";" for splitting sections.

johnml1135:
*[`machine/scripture/parse.py` line 8 at r3](https://reviewable.io/reviews/sillsdev/machine.py/61#-NiuGVDu2e7fwBamliCu:-NiuHqy-D5iYsf0OYE8t:b-68ykto) ([raw file](https://github.com/sillsdev/machine.py/blob/55c982dd528aba15e556a1023a8a171443b2318e/machine/scripture/parse.py#L8)):*

Books can include numbers - such as 1CO.
</blockquote></details>

Nevermind - this is the simple selection, not the advanced.   Ignore my comment.

isaac091:
> The order shouldn't matter - I would keep it as original versification order. Are you accounting for removals as well?

I want to make sure I'm understanding - you're saying to add whole books last? I'm currently doing the removals after everything is added, but not allowing removals of spans of books. I can also change that to allow span removals.

> Nevermind - this is the simple selection, not the advanced. Ignore my comment.

What do you mean by simple vs advanced selection? You were right originally, I need to consider books with numbers in the title.

johnml1135:
Let me try it again: for `MAT-ROM; ACT1-3`, the ACT1-3 would be redundant.  Instead the user should put in `MAT-ROM;-ACT4-28`.  Is that how you understand it?

johnml1135 (commented):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)* status: all files reviewed, 8 unresolved discussions (waiting on @isaac091)

isaac091:
I see. For my current code, `MAT-ROM; ACT1-3` and `MAT-ROM;-ACT4-28` give the same result. Basically, if chapters are specified for a book, the whole book is never added. I did it this way because I thought it would be more convenient for the user so they wouldn't have do something like `MAT-JHN;ACT1-3;ROM`, but I didn't think about doing `MAT-ROM;-ACT4-28`, which is not any more complicated, and probably more intuitive.

ddaspit (commented):
I would just apply them in the order that they appear. That way the precedence is explicit.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61#-:-NiuQR2b3V-nJQVnKnUQ:b-den2ma)* status: all files reviewed, 8 unresolved discussions (waiting on @isaac091)

___
*[`machine/scripture/parse.py` line 77 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu7vuB-1Ili_mztYMZ:-NiuR6W746Zpzqkg0l-o:bbdkler) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L77)):*

We should raise an exception if the chapter is not valid, i.e. out of range.
</blockquote></details>

I would also strip spaces from the chapter numbers.

johnml1135:
I agree with Damein.

johnml1135 (commented):
> Reviewed 2 of 5 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)* status: all files reviewed, 8 unresolved discussions (waiting on @isaac091)

ddaspit (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)* status: all files reviewed, 2 unresolved discussions (waiting on @isaac091 and @johnml1135)

___
*[`machine/scripture/parse.py` line 45 at r2](https://reviewable.io/reviews/sillsdev/machine.py/61#-Niu4Dya7DT_2u7cBVjX:-Njiaxo1C-jK4ob1ZQnM:bjyoctp) ([raw file](https://github.com/sillsdev/machine.py/blob/f9f477d34ba2cf3d4fbd428b355ccf732764bfd4/machine/scripture/parse.py#L45)):*

You should pass in the versification as an argument.
</blockquote></details>

You should use the `ORIGINAL_VERSIFICATION` constant rather than creating a new instance.

johnml1135:
The updated tests look like they cover the cases that I was concerned about.  If they pass, it looks good to me.

johnml1135 (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

ddaspit (commented):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`machine/scripture/__init__.py` line 39 at r5](https://reviewable.io/reviews/sillsdev/machine.py/61#-NjmxwFP74YQ6sdmXO8H:-NjmxwFP74YQ6sdmXO8I:b-dwrspy) ([raw file](https://github.com/sillsdev/machine.py/blob/9f460cfdd502deeecea44b2ba9ba431def2d025e/machine/scripture/__init__.py#L39)):*
> ```Python
> 
> 
> def __getattr__(name: str) -> Any:
> ```

You can remove this function. It is no longer needed.

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/61#-:-Njn5Val1-Gz_lzLwrG1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @isaac091)

Issue #62, opened 2023-11-13
ValueError : 'grc_Cprt' is not a valid language code.


pmachapman:
When attempting to run a pre-translation build for the following translation engine via Serval on QA:

```json
{
  "id": "65525eafbb0376665c83de75",
  "url": "/api/v1/translation/engines/65525eafbb0376665c83de75",
  "name": "65525d45fe7aa96b80ba27bf",
  "sourceLanguage": "grc",
  "targetLanguage": "en",
  "type": "Nmt",
  "isBuilding": false,
  "modelRevision": 0,
  "confidence": 0,
  "corpusSize": 0
}
```

I receive the following error (via the Serval build message):

```
ValueError : 'grc_Cprt' is not a valid language code.
```

grc (Ancient Greek) is the language for several Koine Greek translations in the DBL (grcF35, grcbrent, GRCTR, TCGNT, ...). It looks like `IetfLanguageTag.TryGetSubtags` is returning the script as Cypriot?

johnml1135:
@ddaspit  and @mshannon-sil - Why is this happening?  Is it that we are inferencing only and the language codes are not added?  Is it a bug introduced recently?

Here is a failing experiment, but not exactly the same one: https://app.sil.hosted.allegro.ai/projects/*/experiments/90a7ad732a9d40aa83e2b96ed7836d42/info-output/log?columns=selected&columns=type&columns=name&columns=tags&columns=status&columns=project.name&columns=users&columns=started&columns=last_update&columns=last_iteration&columns=parent.name&order=-started&filter=&deep=true

https://github.com/sillsdev/machine.py/blob/d307a1bde16094927ee7fafca3a588241c94560f/machine/translation/huggingface/hugging_face_nmt_engine.py#L45C1-L57C80

https://github.com/sillsdev/machine.py/blob/d307a1bde16094927ee7fafca3a588241c94560f/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L236-L240

johnml1135:
Yes - it is about inferencing.  Is the error code the best?  Could we say instead "No matching training segments identified.  For inferencing only, both language codes need to be in the NLLB-200 model."

```
2023-11-10 17:27:58,086 - machine.jobs.nmt_engine_build_job - INFO - Downloading data files
2023-11-10 17:27:58,707 - machine.jobs.nmt_engine_build_job - INFO - No matching entries in the source and target corpus - skipping training
[...]

PR #63, opened 2023-11-20
Better error messages for #62

johnml1135:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/63)

### DIFF
--- a/machine/translation/huggingface/hugging_face_nmt_engine.py
+++ b/machine/translation/huggingface/hugging_face_nmt_engine.py
@@ -54,7 +54,7 @@ class HuggingFaceNmtEngine(TranslationEngine):
                 and tgt_lang not in cast(Any, self._tokenizer).lang_code_to_id
                 and tgt_lang not in additional_special_tokens
             ):
-                raise ValueError(f"The specified model does not support the language code '{src_lang}'")
+                raise ValueError(f"The specified model does not support the language code '{tgt_lang}'")
 
         self._pipeline = _TranslationPipeline(
             model=model,
--- a/machine/translation/huggingface/hugging_face_nmt_engine.py
+++ b/machine/translation/huggingface/hugging_face_nmt_engine.py
@@ -47,14 +47,18 @@ class HuggingFaceNmtEngine(TranslationEngine):
                 and src_lang not in cast(Any, self._tokenizer).lang_code_to_id
                 and src_lang not in additional_special_tokens
             ):
-                raise ValueError(f"'{src_lang}' is not a valid language code.")
+                raise ValueError(
+                    f"'{src_lang}' is not a valid language code. This error can happen when there is no matching training data and the language code is not in the NLLB-200."
+                )
 
             if (
                 tgt_lang is not None
                 and tgt_lang not in cast(Any, self._tokenizer).lang_code_to_id
                 and tgt_lang not in additional_special_tokens
             ):
-                raise ValueError(f"'{tgt_lang}' is not a valid language code.")
+                raise ValueError(
+                    f"'{tgt_lang}' is not a valid language code. This error can happen when there is no matching training data and the language code is not in the NLLB-200."
+                )
 
         self._pipeline = _TranslationPipeline(
             model=model,
--- a/machine/translation/huggingface/hugging_face_nmt_engine.py
+++ b/machine/translation/huggingface/hugging_face_nmt_engine.py
@@ -47,20 +47,14 @@ class HuggingFaceNmtEngine(TranslationEngine):
                 and src_lang not in cast(Any, self._tokenizer).lang_code_to_id
                 and src_lang not in additional_special_tokens
             ):
-                raise ValueError(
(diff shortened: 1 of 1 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/63)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`machine/translation/huggingface/hugging_face_nmt_engine.py` line 52 at r1](https://reviewable.io/reviews/sillsdev/machine.py/63#-NjhtAYb4XqtUHXXWjCw:-NjhtAYb4XqtUHXXWjCx:bx2wspt) ([raw file](https://github.com/sillsdev/machine.py/blob/7f7d8e9186d77b9e3220ced9b1f4b5f319999f48/machine/translation/huggingface/hugging_face_nmt_engine.py#L52)):*
> ```Python
>                 raise ValueError(
>                     f"'{src_lang}' is not a valid language code. This error can happen when "
>                     + "there is no matching training data and the language code is not in the NLLB-200."
> ```

The error message is not appropriate in this context. It references "training data", which is specific to the workflow of a build job. It also references NLLB, but this class is not specific to NLLB. There are two options:
1. Update the message here to give a little more detail, i.e. `The specified model does not support the language code '{src_lang}'.` or something similar.
2. Catch the error in `HuggingFaceNmtModelFactory` and throw a new exception with a more detailed message. In that case, we will need to raise a custom error here, so that we can avoid hiding other `ValueError` instances.

mshannon-sil:
I am in favor of the first option, since saying that "The specified model does not support the language code '{src_lang}'." is clearer terminology and implicitly directs the user to adjust the model to support the language code, rather than saying that a particular language code "is not valid" and having the user potentially think the issue is with the language code itself.

johnml1135:
*[`machine/translation/huggingface/hugging_face_nmt_engine.py` line 52 at r1](https://reviewable.io/reviews/sillsdev/machine.py/63#-NjhtAYb4XqtUHXXWjCw:-NjmUIhd3UXJundj4i5c:bkt299x) ([raw file](https://github.com/sillsdev/machine.py/blob/7f7d8e9186d77b9e3220ced9b1f4b5f319999f48/machine/translation/huggingface/hugging_face_nmt_engine.py#L52)):*

The error message is not appropriate in this context. It references "training data", which is specific to the workflow of a build job. It also references NLLB, but this class is not specific to NLLB. There are two options:
1. Update the message here to give a little more detail, i.e. `The specified model does not support the language code '{src_lang}'.` or something similar.
2. Catch the error in `HuggingFaceNmtModelFactory` and throw a new exception with a more detailed message. In that case, we will need to raise a custom error here, so that we can avoid hiding other `ValueError` instances.
</blockquote></details>

Option 1 is good enough.  The error is being raised in production because of another issue that is being fixed now.

johnml1135:
Done.

ddaspit (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/63)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`machine/translation/huggingface/hugging_face_nmt_engine.py` line 57 at r2](https://reviewable.io/reviews/sillsdev/machine.py/63#-NjmpeQq5LemTaxL52V1:-NjmpeQq5LemTaxL52V2:b-7qkl9l) ([raw file](https://github.com/sillsdev/machine.py/blob/cf606611a8c6bd6bb81936b0ddbbf2ae1f2e2bca/machine/translation/huggingface/hugging_face_nmt_engine.py#L57)):*
> ```Python
>                 and tgt_lang not in additional_special_tokens
>             ):
>                 raise ValueError(f"The specified model does not support the language code '{src_lang}'")
> ```

This should be `{tgt_lang}`.

johnml1135 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/63)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`machine/translation/huggingface/hugging_face_nmt_engine.py` line 57 at r2](https://reviewable.io/reviews/sillsdev/machine.py/63#-NjmpeQq5LemTaxL52V1:-NjnI17p1dkYFRhaCjXa:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/cf606611a8c6bd6bb81936b0ddbbf2ae1f2e2bca/machine/translation/huggingface/hugging_face_nmt_engine.py#L57)):*

This should be `{tgt_lang}`.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/63#-:-NjnuhtzA2BnAciP615w:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

PR #82, opened 2023-12-04
Upgrade machine.py production docker container to python 3.11

mshannon-sil:
Upgrading the production docker container to python 3.11. I also changed some of the tests I wrote because when I at first updated all of the poetry packages, a few of them stopped passing. However, I've since reverted most of the packages back to their original versions and only upgraded packages that were necessary to upgrade, but I kept the changes to the tests regardless for maximum compatibility.

### DIFF
--- a/.devcontainer/dockerfile
+++ b/.devcontainer/dockerfile
@@ -1,5 +1,5 @@
 #compatability with Tensorflow 2.6.0 as per https://www.tensorflow.org/install/source#gpu
-ARG PYTHON_VERSION=3.11
+ARG PYTHON_VERSION=3.8
 ARG UBUNTU_VERSION=focal
 ARG POETRY_VERSION=1.6.1
 ARG CUDA_VERSION=11.2.2-cudnn8-runtime-ubuntu20.04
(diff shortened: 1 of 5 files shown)


mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/82)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @johnml1135)

___
*[`tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py` line 145 at r1](https://reviewable.io/reviews/sillsdev/machine.py/82#-NkuPMUh1fRIfydDgU5S:-NkuRidz63cSR5EiafpL:bblauyb) ([raw file](https://github.com/sillsdev/machine.py/blob/61cdb6bf70e5eae5c4da420355d41c3dd54af944/tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py#L145)):*

Why would the test without the `_XX` fail on newer versions of python?  We should support language codes that don't include a script code.
</blockquote></details>

I don't think it was the fact that it was a newer version of python that was making it fail; rather, I had updated transformers without needing to, and the updated version caught that I was using the wrong language codes for this tokenizer. The tokenizer that the tiny-random-nllb model uses is actually the slow MBartTokenizer rather than NllbTokenizerFast, maybe because MBartTokenizer takes a lot less space for testing. That tokenizer does include script codes in all of its default language codes. I'm not sure why it wasn't failing earlier when I was using the wrong language codes for this tokenizer, but I guess that one of the most recent updates just fixed it to work properly and fail if it sees an unrecognized language code. So I don't think we need to make any further changes to these tests.

PR #88, opened 2023-12-13
Restrict build options to only update model hyperparameters

### LINKED ISSUE #86: Restrict build options to only update model hyperparameters
The way that the build options is currently implemented allows a user to override any config parameter, including `data_dir` and `model_type`. We should restrict what parameters can be overridden. The build options should be merged with the model settings, i.e. the `huggingface` settings object.

ddaspit:
- fixes #86

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/88)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,26 +18,46 @@
       "request": "launch",
       "module": "machine.jobs.build_nmt_engine",
       "justMyCode": false,
-      "args": [
-        "--model-type",
-        "huggingface",
-        "--build-id",
-        "build1",
-        "--engine-id",
-        "engine1",
-        "--src-lang",
-        "spa_Latn",
-        "--trg-lang",
-        "eng_Latn",
-        "--clearml",
-        "--build-options",
-        "{\"max_steps\": 10}"
-      ]
+      "windows": {
+        "args": [
+          "--model-type",
+          "huggingface",
+          "--build-id",
+          "build1",
+          "--engine-id",
+          "engine1",
+          "--src-lang",
+          "spa_Latn",
+          "--trg-lang",
+          "eng_Latn",
+          "--build-options",
+          "{\\\"train_params\\\": {\\\"max_steps\\\": 10}}"
+        ]
+      },
+      "linux": {
+        "args": [
+          "--model-type",
+          "huggingface",
+          "--build-id",
+          "build1",
+          "--engine-id",
+          "engine1",
+          "--src-lang",
+          "spa_Latn",
+          "--trg-lang",
+          "eng_Latn",
+          "--build-options",
+          "{\"train_params\": {\"max_steps\": 10}}"
+        ]
+      }
     },
     {
-      "name": "Debug Unit Test",
+      "name": "Python: Debug Tests",
       "type": "python",
-      "request": "test",
+      "request": "launch",
+      "program": "${file}",
+      "purpose": ["debug-test"],
+      "console": "integratedTerminal",
       "justMyCode": false
     }
   ]
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,13 +1,12 @@
 {
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": true
+    "source.organizeImports": "explicit"
   },
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true,
   "[python]": {
     "editor.defaultFormatter": "ms-python.black-formatter",
     "editor.formatOnSave": true
-  },
-  "python.formatting.provider": "none"
+  }
 }
--- a/machine/jobs/build_nmt_engine.py
+++ b/machine/jobs/build_nmt_engine.py
@@ -45,20 +45,20 @@ def run(args: dict) -> None:
         logger.info("NMT Engine Build Job started")
 
         SETTINGS.update(args)
-        try:
-            SETTINGS.build_options = json.loads(args["build_options"])
-        except ValueError as e:
-            raise ValueError("Build options could not be parsed: Invalid JSON") from e
-        except TypeError as e:
-            raise TypeError(f"Build options could not be parsed: {e}") from e
-        if SETTINGS.build_options:
-            SETTINGS.update(SETTINGS.build_options)
+        model_type = cast(str, SETTINGS.model_type).lower()
+        if "build_options" in SETTINGS:
+            try:
+                build_options = json.loads(cast(str, SETTINGS.build_options))
+            except ValueError as e:
+                raise ValueError("Build options could not be parsed: Invalid JSON") from e
+            except TypeError as e:
+                raise TypeError(f"Build options could not be parsed: {e}") from e
+            SETTINGS.update({model_type: build_options})
         SETTINGS.data_dir = os.path.expanduser(cast(str, SETTINGS.data_dir))
 
         logger.info(f"Config: {SETTINGS.as_dict()}")
 
         shared_file_service = ClearMLSharedFileService(SETTINGS)
-        model_type = cast(str, SETTINGS.model_type).lower()
         nmt_model_factory: NmtModelFactory
         if model_type == "huggingface":
             from .huggingface.hugging_face_nmt_model_factory import HuggingFaceNmtModelFactory
@@ -87,7 +87,7 @@ def main() -> None:
     parser.add_argument("--src-lang", required=True, type=str, help="Source language tag")
     parser.add_argument("--trg-lang", required=True, type=str, help="Target language tag")
     parser.add_argument("--clearml", default=False, action="store_true", help="Initializes a ClearML task")
-    parser.add_argument("--build-options", default="{}", type=str, help="Build configurations")
+    parser.add_argument("--build-options", default=None, type=str, help="Build configurations")
     args = parser.parse_args()
 
     run({k: v for k, v in vars(args).items() if v is not None})
--- a/machine/jobs/huggingface/hugging_face_nmt_model_factory.py
+++ b/machine/jobs/huggingface/hugging_face_nmt_model_factory.py
@@ -26,12 +26,13 @@ class HuggingFaceNmtModelFactory(NmtModelFactory):
         args = config.huggingface.train_params.to_dict()
         args["output_dir"] = str(self._model_dir)
         args["overwrite_output_dir"] = True
-        if "max_steps" in self._config:
-            if self._config.max_steps > 50000:
-                raise ValueError("max_steps must be less than or equal to 50000")
-            args["max_steps"] = self._config.max_steps
+        # Use "max_steps" from root for backward compatibility
+        if "max_steps" in self._config.huggingface:
+            args["max_steps"] = self._config.huggingface.max_steps
         parser = HfArgumentParser(cast(Any, Seq2SeqTrainingArguments))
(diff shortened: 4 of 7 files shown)


johnml1135:
*[`machine/jobs/settings.yaml` line 4 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_hzJ-3bCr0Go86Vpj:-Nl_hzJ-3bCr0Go86Vpk:b8bypx) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/settings.yaml#L4)):*
> ```YAML
>   model_type: huggingface
>   data_dir: ~/machine
>   pretranslation_batch_size: 1024
> ```

So, batch size is really pretranslation_batch_size, which is the size of corpus we try to pretranslate at the same time.  You don't want this to be editable by the user, hence not putting it in huggingface.

johnml1135:
*[`machine/jobs/settings.yaml` line 18 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_iCH-0JCXlvOfqwcQ:-Nl_iCH-0JCXlvOfqwcR:b-fxcgtc) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/settings.yaml#L18)):*
> ```YAML
>       fp16: true
>       save_strategy: no
>       max_steps: 20000
> ```

max_steps goes under train_params, which is more appropriate.  I agree.

johnml1135:
*[`machine/jobs/huggingface/hugging_face_nmt_model_factory.py` line 31 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_imwK64KdvThHtgtI:-Nl_imwK64KdvThHtgtJ:bv0xmw1) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/huggingface/hugging_face_nmt_model_factory.py#L31)):*
> ```Python
>         # Use "max_steps" from root for backward compatibility
>         if "max_steps" in self._config.huggingface:
>             args["max_steps"] = self._config.huggingface.max_steps
> ```

So, let's see if I follow this logic.  a person can set max_steps from the root and it will work successfully, but the real place is to set it in `huggingface.tran_params.max_steps`.  The issue is that you took my updates at looking for max_steps in `self._config.huggingface` instead of just `self._config`.  Am I missing something?

johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_jdF72X8CtQlId_bU:-Nl_jdF72X8CtQlId_bV:bw0psvx) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/build_nmt_engine.py#L48)):*
> ```Python
> 
>         SETTINGS.update(args)
>         model_type = cast(str, SETTINGS.model_type).lower()
> ```

So, let's see if I get this.  In settings.yml, there is a model_type parameter, here Huggingface.  That is taken from settings.yml without being tampered by any input arguments.  Then, any `build_options` given to Serval end up going under the current `model_type`, as seen a few lines below.  This seems sensible enough, except that, why not rename `huggingface` to `build_options` in settings.yml?  The `model_type` is already specified and it will be more clear in the documentation why the `build_options` map to ... `build_options`.

johnml1135:
*[`machine/jobs/huggingface/hugging_face_nmt_model_factory.py` line 31 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_imwK64KdvThHtgtI:-Nl_lqZBCwV1BBnWxvq_:b-hs4x17) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/huggingface/hugging_face_nmt_model_factory.py#L31)):*

So, let's see if I follow this logic.  a person can set max_steps from the root and it will work successfully, but the real place is to set it in `huggingface.tran_params.max_steps`.  The issue is that you took my updates at looking for max_steps in `self._config.huggingface` instead of just `self._config`.  Am I missing something?
</blockquote></details>

I read over the import of the settings and I see what you are doing now.  My main concern is the above concern to rename the `huggingface` group just `build_options`.

johnml1135:
*[`machine/jobs/nmt_engine_build_job.py` line 89 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_mHlc8dYMTpkNBsMz:-Nl_mHlc8dYMTpkNBsN-:bec6xzr) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/nmt_engine_build_job.py#L89)):*
> ```Python
>             current_inference_step = 0
>             phase_progress(ProgressStatus.from_step(current_inference_step, inference_step_count))
>             batch_size = self._config["pretranslation_batch_size"]
> ```

If `pretranslation_batch_size` will not be changed in any meaningful way for any meaningful reason, why include it?  It isn't the batch size that the GPU uses for doing Pretranslations but the number of segments that are pulled from the source and handled at one time.  Why not just set a constant to 1024 and be done with it?

johnml1135:
*[`tests/jobs/test_nmt_engine_build_job.py` line 37 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_mltDEukSBUB4UzbB:-Nl_mltE2tUA6F1_zC0l:b-gugoga) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/tests/jobs/test_nmt_engine_build_job.py#L37)):*
> ```Python
> class _TestEnvironment:
>     def __init__(self, decoy: Decoy) -> None:
>         config = {"src_lang": "es", "trg_lang": "en", "pretranslation_batch_size": 100}
> ```

Why would this need to be 100 here?  (see other comment on `pretranslation_batch_size`

johnml1135 (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/88)* status: all files reviewed, 5 unresolved discussions (waiting on @ddaspit)

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/88)* status: all files reviewed, 5 unresolved discussions (waiting on @johnml1135)

___
*[`machine/jobs/build_nmt_engine.py` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_jdF72X8CtQlId_bU:-NldR0ce03viuJrWd1_9:b-g7laqg) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/build_nmt_engine.py#L48)):*

So, let's see if I get this.  In settings.yml, there is a model_type parameter, here Huggingface.  That is taken from settings.yml without being tampered by any input arguments.  Then, any `build_options` given to Serval end up going under the current `model_type`, as seen a few lines below.  This seems sensible enough, except that, why not rename `huggingface` to `build_options` in settings.yml?  The `model_type` is already specified and it will be more clear in the documentation why the `build_options` map to ... `build_options`.
</blockquote></details>

If we support multiple model types, we can include the parameters for the different model types in the same settings file.
___
*[`machine/jobs/nmt_engine_build_job.py` line 89 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_mHlc8dYMTpkNBsMz:-NldRRlTDP5pxJwmny0P:bkbmz0g) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/nmt_engine_build_job.py#L89)):*

If `pretranslation_batch_size` will not be changed in any meaningful way for any meaningful reason, why include it?  It isn't the batch size that the GPU uses for doing Pretranslations but the number of segments that are pulled from the source and handled at one time.  Why not just set a constant to 1024 and be done with it?
</blockquote></details>

It could be a constant. I just made it a setting, so it would be slightly easier to tweak if we need to.
___
*[`machine/jobs/settings.yaml` line 4 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_hzJ-3bCr0Go86Vpj:-NldRqyZ2DD3S1AdkyHB:b-ew7aaa) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/settings.yaml#L4)):*

So, batch size is really pretranslation_batch_size, which is the size of corpus we try to pretranslate at the same time.  You don't want this to be editable by the user, hence not putting it in huggingface.
</blockquote></details>

That is correct. This is the number of source segments to buffer while generating translations. It is different than the training or inferencing batch size.
___
*[`tests/jobs/test_nmt_engine_build_job.py` line 37 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_mltDEukSBUB4UzbB:-NldrJay88P6Irj3xzp3:b-q7wv6z) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/tests/jobs/test_nmt_engine_build_job.py#L37)):*

Why would this need to be 100 here?  (see other comment on `pretranslation_batch_size`
</blockquote></details>

It is mostly an arbitrary number for testing.

johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_jdF72X8CtQlId_bU:-Nle2ofuEINzoMW8CWE6:b-82mjdc) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/build_nmt_engine.py#L48)):*

If we support multiple model types, we can include the parameters for the different model types in the same settings file.
</blockquote></details>

I could go either way - ok.

johnml1135:
*[`machine/jobs/nmt_engine_build_job.py` line 89 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_mHlc8dYMTpkNBsMz:-Nle3CpK0CMYxshIWLA-:b2ni9bh) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/machine/jobs/nmt_engine_build_job.py#L89)):*

It could be a constant. I just made it a setting, so it would be slightly easier to tweak if we need to.
</blockquote></details>

ok.

johnml1135:
*[`tests/jobs/test_nmt_engine_build_job.py` line 37 at r1](https://reviewable.io/reviews/sillsdev/machine.py/88#-Nl_mltDEukSBUB4UzbB:-Nle3Nkh7ydznAnePQgc:b2ni9bh) ([raw file](https://github.com/sillsdev/machine.py/blob/8b9234e858ee2abd6b9d0601ad8eb4b3be246237/tests/jobs/test_nmt_engine_build_job.py#L37)):*

It is mostly an arbitrary number for testing.
</blockquote></details>

ok.

johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/88)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit)

PR #97, opened 2024-01-23
Clean up get_chapters code and expand tests

isaac091:
@Enkidu93 For the first section, I thought my original error messages didn't make a ton of sense, so I rewrote and combined them. I also updated the `BOOK_RANGE` and `CHAPTER_SELECTION` regexes to allow subtraction so that I could take out the subtraction check right before.

I also decided to just take out the ability to add "chapter 0" because I don't think anyone's using it and it's inconsistent since adding the whole book doesn't include chapter 0, like we talked about.

I kept whole books being represented as an empty list because it's much more convenient that way for silnlp. Otherwise, we'd have to pass a versification variable through all the different functions just to be able to check chapter counts.

Other than that, I fixed the stuff for backwards ranges and added your test cases. Let me know if you have more thoughts or if I missed something!

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/97)

### DIFF
--- a/machine/scripture/parse.py
+++ b/machine/scripture/parse.py
@@ -1,13 +1,13 @@
 import re
 from typing import Dict, List, Set, Union
 
-from .canon import book_id_to_number
+from .canon import book_id_to_number, book_number_to_id
 from .constants import ORIGINAL_VERSIFICATION
 from .verse_ref import Versification
 
 COMMA_SEPARATED_BOOKS = re.compile(r"([A-Z\d]{3}|OT|NT)(, ?([A-Z\d]{3}|OT|NT))*")
-BOOK_RANGE = re.compile(r"[A-Z\d]{3}-[A-Z\d]{3}")
-CHAPTER_SELECTION = re.compile(r"[A-Z\d]{3} ?(\d+|\d+-\d+)(, ?(\d+|\d+-\d+))*")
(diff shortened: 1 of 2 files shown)


ddaspit (approved):
Paratext does have the concept of chapter 0. It includes everything before the first chapter marker. Scripture Forge doesn't have the same concept. Chapter 1 includes all of the introductory material. I guess we should be consistent with Scripture Forge, so it should be okay to remove support for chapter 0.

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/97#-:-NosWNBk389TP_4K1gPP:bex5ws5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93 (approved):
Excellent! Thank you, Isaac, for the quick turnaround. I'll make sure the C# matches now after your additional changes and add matching tests.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/97#-:-NovqVgcEImRkzVHJjbG:b-5vzc7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #99, opened 2024-02-02
Add option to save the model during build job

ddaspit:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,6 @@
   "[python]": {
     "editor.defaultFormatter": "ms-python.black-formatter",
     "editor.formatOnSave": true
-  }
+  },
+  "black-formatter.path": ["poetry", "run", "black"]
 }
--- a/machine/jobs/build_nmt_engine.py
+++ b/machine/jobs/build_nmt_engine.py
@@ -88,6 +88,7 @@ def main() -> None:
     parser.add_argument("--trg-lang", required=True, type=str, help="Target language tag")
     parser.add_argument("--clearml", default=False, action="store_true", help="Initializes a ClearML task")
     parser.add_argument("--build-options", default=None, type=str, help="Build configurations")
+    parser.add_argument("--save-model", default=None, type=str, help="Save the model using the specified base name")
     args = parser.parse_args()
 
     run({k: v for k, v in vars(args).items() if v is not None})
--- a/machine/jobs/huggingface/hugging_face_nmt_model_factory.py
+++ b/machine/jobs/huggingface/hugging_face_nmt_model_factory.py
@@ -1,4 +1,5 @@
 import logging
+import tarfile
 from pathlib import Path
 from typing import Any, cast
 
@@ -84,7 +85,15 @@ class HuggingFaceNmtModelFactory(NmtModelFactory):
         )
 
     def save_model(self) -> None:
-        self._shared_file_service.save_model(self._model_dir)
+        if self._config.save_model is None:
+            return
+
+        tar_file_path = Path(self._config.data_dir, "builds", self._config.build_id, "model.tar.gz")
+        with tarfile.open(tar_file_path, "w:gz") as tar:
+            for path in self._model_dir.iterdir():
+                if path.is_file():
+                    tar.add(path, arcname=path.name)
+        self._shared_file_service.save_model(tar_file_path, self._config.save_model + ".tar.gz")
 
     @property
     def _model_dir(self) -> Path:
--- a/machine/jobs/huggingface/hugging_face_nmt_model_factory.py
+++ b/machine/jobs/huggingface/hugging_face_nmt_model_factory.py
@@ -85,7 +85,7 @@ class HuggingFaceNmtModelFactory(NmtModelFactory):
         )
 
     def save_model(self) -> None:
-        if self._config.save_model is None:
+        if "save_model" not in self._config:
             return
 
         tar_file_path = Path(self._config.data_dir, "builds", self._config.build_id, "model.tar.gz")
--- a/machine/jobs/nmt_engine_build_job.py
+++ b/machine/jobs/nmt_engine_build_job.py
@@ -94,6 +94,10 @@ class NmtEngineBuildJob:
                 current_inference_step += len(pi_batch)
                 phase_progress(ProgressStatus.from_step(current_inference_step, inference_step_count))
 
+        if self._config.save_model is not None:
+            logger.info("Saving model")
+            self._nmt_model_factory.save_model()
+
 
 def _translate_batch(
     engine: TranslationEngine,
--- a/machine/jobs/nmt_engine_build_job.py
+++ b/machine/jobs/nmt_engine_build_job.py
@@ -94,7 +94,7 @@ class NmtEngineBuildJob:
                 current_inference_step += len(pi_batch)
                 phase_progress(ProgressStatus.from_step(current_inference_step, inference_step_count))
 
-        if self._config.save_model is not None:
+        if "save_model" in self._config and self._config.save_model is not None:
             logger.info("Saving model")
             self._nmt_model_factory.save_model()
 
--- a/machine/jobs/shared_file_service.py
+++ b/machine/jobs/shared_file_service.py
@@ -76,8 +76,11 @@ class SharedFileService(ABC):
     def get_parent_model(self, language_tag: str) -> Path:
         return self._download_folder(f"parent_models/{language_tag}", cache=True)
 
-    def save_model(self, model_dir: Path) -> None:
-        self._upload_folder(f"models/{self._engine_id}", model_dir)
+    def save_model(self, model_path: Path, name: str) -> None:
+        if model_path.is_file():
+            self._upload_file(f"models/{name}", model_path)
+        else:
+            self._upload_folder(f"models/{name}", model_path)
 
     @property
     def _data_dir(self) -> Path:


johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 91 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpdlYwn8BGHaC9IvLgw:-NpdlYwn8BGHaC9IvLgx:b-e2ckqr) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/build_nmt_engine.py#L91)):*
> ```Python
>     parser.add_argument("--clearml", default=False, action="store_true", help="Initializes a ClearML task")
>     parser.add_argument("--build-options", default=None, type=str, help="Build configurations")
>     parser.add_argument("--save-model", default=False, action="store_true", help="Save the model")
> ```

So we would be able to invoke it through a command line option or by putting `{save-model: true}` in the build-options?

johnml1135:
*[`machine/jobs/shared_file_service.py` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-Npdlrtp-A6nLX6A6Hiz:-Npdlrtp-A6nLX6A6Hj-:b-clgj9h) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/shared_file_service.py#L81)):*
> ```Python
>     def save_model(self, model_path: Path) -> None:
>         if model_path.is_file():
>             self._upload_file(f"models/{self._build_id}" + "".join(model_path.suffixes), model_path)
> ```

Wasn't the naming going to be `{engineId}_{buildRevision}`?  Otherwise, the machine side would be much more complicated.  I can send down the buildRevision as a passed argument.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`machine/jobs/build_nmt_engine.py` line 91 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpdlYwn8BGHaC9IvLgw:-NpeFCabAYmv1ezUn0Hm:b-3okfvg) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/build_nmt_engine.py#L91)):*

So we would be able to invoke it through a command line option or by putting `{save-model: true}` in the build-options?
</blockquote></details>

You can pass it as an arg in the job script. Putting it in the build options won't work. I am assuming that the parameter to save a model would be specified in the build config.
___
*[`machine/jobs/shared_file_service.py` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-Npdlrtp-A6nLX6A6Hiz:-NpeG-Y5DM5rVqZBdRIJ:b1mj7jd) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/shared_file_service.py#L81)):*

Wasn't the naming going to be `{engineId}_{buildRevision}`?  Otherwise, the machine side would be much more complicated.  I can send down the buildRevision as a passed argument.
</blockquote></details>

Yes, but I thought about it a bit more. The model download would be associated with the build from the Serval perspective. So Serval will request a download url from the engine using the build id. If the file is labelled with the build id, Machine should be able to easily for the download url.

johnml1135:
*[`.vscode/settings.json` line 12 at r2](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpeNNRAD7PA4PBvUJqh:-NpeNNRAD7PA4PBvUJqi:bq2hqic) ([raw file](https://github.com/sillsdev/machine.py/blob/89179812bfb3e91443fba1488fbdb80cfb2e0388/.vscode/settings.json#L12)):*
> ```JSON
>     "editor.formatOnSave": true
>   },
>   "black-formatter.path": ["poetry", "run", "black"]
> ```

the back formatter is already called out - is this to make it call poetry's version instead of whatever is running in the command line?

johnml1135:
*[`machine/jobs/build_nmt_engine.py` line 91 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpdlYwn8BGHaC9IvLgw:-NpeNbOm5s03GPArGbvv:b-jsqann) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/build_nmt_engine.py#L91)):*

You can pass it as an arg in the job script. Putting it in the build options won't work. I am assuming that the parameter to save a model would be specified in the build config.
</blockquote></details>

Good enough for me.

johnml1135:
*[`machine/jobs/shared_file_service.py` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-Npdlrtp-A6nLX6A6Hiz:-NpeNmOF00GIU3hossza:b-u2fsft) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/shared_file_service.py#L81)):*

Yes, but I thought about it a bit more. The model download would be associated with the build from the Serval perspective. So Serval will request a download url from the engine using the build id. If the file is labelled with the build id, Machine should be able to easily for the download url.
</blockquote></details>

A few things:
* What if the most recent build failed or wasn't saved?  Then Serval would have to implement logic as to which build was valid and saved and is now requestd
* As for removing old models, since only the most recent one could be requested, I am wanting to delete old models from previous builds.  Using the engineId_buildNumber scheme would make that straightforward.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @johnml1135)

___
*[`machine/jobs/shared_file_service.py` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-Npdlrtp-A6nLX6A6Hiz:-NpeOkxVEdSUxrigTvWx:bhcxa1f) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/shared_file_service.py#L81)):*

A few things:
* What if the most recent build failed or wasn't saved?  Then Serval would have to implement logic as to which build was valid and saved and is now requestd
* As for removing old models, since only the most recent one could be requested, I am wanting to delete old models from previous builds.  Using the engineId_buildNumber scheme would make that straightforward.
</blockquote></details>

- When a download url is requested for a build's model, Serval would be able to check if the build was successful and a model was saved. If the model isn't available for that build, Serval would return a 4xx status code, otherwise it would request the download url from the engine.
- I thought that we could just remove the old models based on the date created/modified. We probably don't want to delete a model as soon as another model is trained for the same engine. A user might be in the middle of downloading the model.

johnml1135 (commented):
Do we know why CodeCov is not working on these branches?

> Reviewed 4 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99#-:-NpeO9bmBosoA23DXQEU:b-nptovf)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

johnml1135:
*[`machine/jobs/shared_file_service.py` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine.py/99#-Npdlrtp-A6nLX6A6Hiz:-NpeXXkn59AKUhbjNQ50:b-rkprjj) ([raw file](https://github.com/sillsdev/machine.py/blob/65bc4d49688a94b01348ed67f81ccce76e01a9c6/machine/jobs/shared_file_service.py#L81)):*

- When a download url is requested for a build's model, Serval would be able to check if the build was successful and a model was saved. If the model isn't available for that build, Serval would return a 4xx status code, otherwise it would request the download url from the engine.
- I thought that we could just remove the old models based on the date created/modified. We probably don't want to delete a model as soon as another model is trained for the same engine. A user might be in the middle of downloading the model.
</blockquote></details>

* For Killing old models, I was thinking of marking it twice and then killing it after a day if it is not needed, to address issues like that.
* IDX was concerned about deleting models after 30 days - that they may want to use them for longer term things.

johnml1135:
It isn't working because of the OutOfMemory and torch 2.1.2 issue.  I'll ignore coverage for the time being.

johnml1135:
*[`machine/jobs/shared_file_service.py` line 79 at r3](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpjDVga707ihM2PGZ18:-NpjDVga707ihM2PGZ19:bzcnbon) ([raw file](https://github.com/sillsdev/machine.py/blob/25b93ac48bccdf28d0a78b86593d893fd09da7a7/machine/jobs/shared_file_service.py#L79)):*
> ```Python
>         return self._download_folder(f"parent_models/{language_tag}", cache=True)
> 
>     def save_model(self, model_path: Path, name: str) -> None:
> ```

What about the suffixes?

johnml1135:
*[`machine/jobs/shared_file_service.py` line 79 at r3](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpjDVga707ihM2PGZ18:-NpjDarg5sXW6JmPStnh:b6ulc76) ([raw file](https://github.com/sillsdev/machine.py/blob/25b93ac48bccdf28d0a78b86593d893fd09da7a7/machine/jobs/shared_file_service.py#L79)):*

What about the suffixes?
</blockquote></details>

Ah - I see it now.

johnml1135 (commented):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

johnml1135:
*[`.vscode/settings.json` line 12 at r2](https://reviewable.io/reviews/sillsdev/machine.py/99#-NpeNNRAD7PA4PBvUJqh:-NpjDhSt8sf2vmxnhvwZ:b-2u3u2n) ([raw file](https://github.com/sillsdev/machine.py/blob/89179812bfb3e91443fba1488fbdb80cfb2e0388/.vscode/settings.json#L12)):*

the back formatter is already called out - is this to make it call poetry's version instead of whatever is running in the command line?
</blockquote></details>

Not worth holding up.  It should be fine.

johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135 (approved):
> Reviewed 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/99)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135:
There's code cov!

johnml1135:
:lgtm:

PR #100, opened 2024-02-02
And upgrade torch, numpy, pandas and accelerate as well as black, pyr…

johnml1135:
…ight and flake8.  To fix the OutOfMemory Error, but also to keep current.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/100)


johnml1135:
All E2E tests passed - this should be ok.

Issue #102, opened 2024-03-27
Port Machine's USFM generation changes to machine.py


mshannon-sil:
Changes to USFM generation in Machine from https://github.com/sillsdev/machine/pull/160 need to be ported to machine.py.

mshannon-sil:
Changes have already been ported from [sillsdev/machine#160](https://github.com/sillsdev/machine/pull/160) and [sillsdev/machine#163](https://github.com/sillsdev/machine/pull/163). However, there have been more USFM related changes in Machine since then that also need to be ported over.

PR #111, opened 2024-07-19
Port USFM code from Machine #102

mshannon-sil:


ddaspit (commented):
Is this ready to be reviewed? Does this include all of the commits that need to be ported?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111#-:-O2VBvU87vUQqKlT7QiJ:b-ewyxh2)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @mshannon-sil)

mshannon-sil (commented):
Not yet, currently finishing up the commit for non-verse text support. Should be done with that by the end of tomorrow. There's still more commits to port after that, but if you'd like to review the PR midway through the porting process, that could be a good point to do so.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111#-:-O2VqEQB44bQa3SZ96i5:bursr4l)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @mshannon-sil)

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111)* status: 0 of 35 files reviewed, 7 unresolved discussions

___
*[`machine/corpora/scripture_ref_usfm_parser_handler.py` line 23 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C3fQLCU4jGMn4QSwk:-O4C3fQLCU4jGMn4QSwl:b-csuwr) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_ref_usfm_parser_handler.py#L23)):*
> ```Python
>     def __init__(self) -> None:
>         self._cur_verse_ref: VerseRef = VerseRef()
>         self._cur_elements_stack: List[ScriptureElement] = []
> ```

For variables that were a Stack type in C#, I used a list in python, since it's fast to append and pop items from a list like a stack. I added `_stack` to the end of these variables to document that they should be used as a stack.
___
*[`machine/corpora/scripture_ref_usfm_parser_handler.py` line 173 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C56lNA3qmtzqJ-yT5:-O4C56lNA3qmtzqJ-yT6:b-sve24m) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_ref_usfm_parser_handler.py#L173)):*
> ```Python
>         )
>         # No need to reverse unlike in Machine, elements are already added in correct order
>         path = [e for e in self._cur_elements_stack if e.position > 0]
> ```

In Machine, there's a need to reverse the order of the stack based on how elements are added to the stack. However when I ported it to machine.py and used a list, all the elements were actually being added in the correct order, and reversing just caused bugs, so I removed that in machine.py.
___
*[`machine/corpora/scripture_text.py` line 43 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C5fRC-jDatworvC_2:-O4C5fRC-jDatworvC_3:bv6xnei) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text.py#L43)):*
> ```Python
>         else:
>             yield from self._create_rows_scripture_ref(ref, text, is_sentence_start)
> ```

In Machine, this method was overloaded to have two separate implementations for either a VerseRef or a List of ScriptureRef as input. Since I can't do exactly the same thing in python, I created a function with the general `create_rows` name, which checks the type of the input and calls the corresponding `create_rows` method for that type.
___
*[`machine/corpora/text_corpus.py` line 203 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C-niL2Jz6-XVRzorP:-O4C-niL2Jz6-XVRzorQ:bncgnnm) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/text_corpus.py#L203)):*
> ```Python
> 
> 
> class _FilterTextCorpus(TextCorpus):
> ```

The class _FilterTextCorpus in machine.py is called WhereTextCorpus in Machine. I thought this might be due to different naming conventions in python vs c#, but then _TextFilterTextCorpus in machine.py has the same name in Machine. Is the naming difference here intentional?
___
*[`machine/scripture/scripture_ref.py` line 18 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C1XrA52lfXCyPQQgy:-O4C1XrA52lfXCyPQQgz:b-wjmnvg) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L18)):*
> ```Python
>         self._path: List[ScriptureElement] = path if path is not None else []
> 
>     _empty: Optional[ScriptureRef] = None
> ```

I was concerned about properties of the `_empty` ScriptureRef (verse_ref and path) getting modified and affecting all instances so that they are no longer empty. But since the public properties only have getters not setters, it should be fine as long as the corresponding private properties aren't modified.
___
*[`machine/scripture/verse_ref.py` line 380 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8ygt7zsh7lzn1xtp:-O4C8ygt7zsh7lzn1xtq:b5acflt) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L380)):*
> ```Python
>         return 0
> 
>     def __eq__(self, other: object) -> bool:
> ```

The Equals method in SIL.Scripture/VerseRef.cs is the same as the exact_equals method that was in machine.py. I changed the machine.py version to be the __eq__ method instead so that it is called when == is used, since this seemed to me to be what we would want, and it didn't break any tests. Let me know if there's a reason to not make it the default equality comparison.
___
*[`machine/scripture/verse_ref.py` line 488 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8LuK1HCjoOcCCDPH:-O4C8LuLAl8CNFAcrEK4:b-k0f5ip) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L488)):*
> ```Python
> 
> def are_overlapping_verse_ranges_vref(verse_ref1: VerseRef, verse_ref2: VerseRef) -> bool:
>     if verse_ref1.is_default or verse_ref2.is_default:
> ```

I think there's actually a bug in SIL.Scripture/VerseRef.cs. The corresponding section in OverlappingVersesRanges for the VerseRef type input is `if (verseRef1.IsDefault || verseRef1.IsDefault)` but it should be `if (verseRef1.IsDefault || verseRef2.IsDefault)`.

ddaspit (changes requested):
> Reviewed 1 of 1 files at r5, 34 of 34 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111)* status: all files reviewed, 8 unresolved discussions (waiting on @mshannon-sil)

___
*[`machine/corpora/scripture_ref_usfm_parser_handler.py` line 23 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C3fQLCU4jGMn4QSwk:-O4GsIqE-ks7uL7sk0TL:b-a7ljwd) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_ref_usfm_parser_handler.py#L23)):*

For variables that were a Stack type in C#, I used a list in python, since it's fast to append and pop items from a list like a stack. I added `_stack` to the end of these variables to document that they should be used as a stack.
</blockquote></details>

Makes sense to me.
___
*[`machine/corpora/scripture_text.py` line 43 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C5fRC-jDatworvC_2:-O4GXtQm-A1QABw19q8J:b-w8o72l) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text.py#L43)):*

In Machine, this method was overloaded to have two separate implementations for either a VerseRef or a List of ScriptureRef as input. Since I can't do exactly the same thing in python, I created a function with the general `create_rows` name, which checks the type of the input and calls the corresponding `create_rows` method for that type.
</blockquote></details>

That is a good solution.
___
*[`machine/corpora/scripture_text.py` line 82 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4G_gqn-DCe8y5VGXrR:-O4G_gqn-DCe8y5VGXrS:b6g5ni8) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text.py#L82)):*
> ```Python
>             )
> 
>     def _create_row(
> ```

Because Python doesn't support overloading, we sometimes have to deviate from the C# code a bit, as you discovered for the `_create_rows` method. I think it would be good to name this something else, maybe `_create_scripture_row`, then you wouldn't have to call the base class `_create_row` method using `super()`. If you rename this method, then you should also rename the `_create_rows` method to match.
___
*[`machine/corpora/scripture_text_corpus.py` line 111 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GeBLIB3IMr3nHrkon:-O4GeBLIB3IMr3nHrkoo:b-30wiw8) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text_corpus.py#L111)):*
> ```Python
> 
> 
> def is_scripture(text_corpus: TextCorpus) -> bool:
> ```

We want to export this function from the `__init__.py` of the `corpora` package.
___
*[`machine/corpora/standard_parallel_text_corpus.py` line 255 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GbzFM3V9bl-0Fo0vW:-O4GbzFM3V9bl-0Fo0vX:bfmomk) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/standard_parallel_text_corpus.py#L255)):*
> ```Python
>         trg_refs = [] if trg_row is None else [trg_row.ref]
> 
>         if len(trg_refs) == 0 and isinstance(self._target_corpus, ScriptureTextCorpus):
> ```

This was a bug in the C# code, you should use the `is_scripture` function.
___
*[`machine/corpora/text_corpus.py` line 203 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C-niL2Jz6-XVRzorP:-O4GWUm-5coYDzQcYXmL:bsmujzy) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/text_corpus.py#L203)):*

The class _FilterTextCorpus in machine.py is called WhereTextCorpus in Machine. I thought this might be due to different naming conventions in python vs c#, but then _TextFilterTextCorpus in machine.py has the same name in Machine. Is the naming difference here intentional?
</blockquote></details>

As you guessed, this is because of the difference in function names between Python and C#. Python uses `map` and `filter`. C# uses `Select` and `Where`. I wanted to match the standard names in each language.
___
*[`machine/scripture/scripture_element.py` line 10 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GnqNM0JtJke216huW:-O4GnqNM0JtJke216huX:b-zhs63j) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_element.py#L10)):*
> ```Python
> 
> @total_ordering
> class ScriptureElement(Comparable):
> ```

This class should be moved to the `corpora` package.  This class should also be exported from `__init__.py` in the `corpora` package.
___
*[`machine/scripture/scripture_ref.py` line 13 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4Gla2142SMa4u_R7TO:-O4Gla2142SMa4u_R7TP:bouzz2m) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L13)):*
> ```Python
> 
> @total_ordering
> class ScriptureRef(Comparable):
> ```

This class should be moved to the `corpora` package.
___
*[`machine/scripture/scripture_ref.py` line 18 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C1XrA52lfXCyPQQgy:-O4GmCVA33JK-gIyODA2:b411ynm) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L18)):*

I was concerned about properties of the `_empty` ScriptureRef (verse_ref and path) getting modified and affecting all instances so that they are no longer empty. But since the public properties only have getters not setters, it should be fine as long as the corresponding private properties aren't modified.
</blockquote></details>

You can make this a top-level constant, i.e. `EMPTY_SCRIPTURE_REF`. 
___
*[`machine/scripture/verse_ref.py` line 380 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8ygt7zsh7lzn1xtp:-O4GqdHg3Ez5Qbj9_H6_:bvcfcal) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L380)):*

The Equals method in SIL.Scripture/VerseRef.cs is the same as the exact_equals method that was in machine.py. I changed the machine.py version to be the __eq__ method instead so that it is called when == is used, since this seemed to me to be what we would want, and it didn't break any tests. Let me know if there's a reason to not make it the default equality comparison.
</blockquote></details>

The `Comparable` base class provides an implementation of `__eq__` based on the `compare_to` method. This makes all of the comparison methods consistent with each other. `exact_equals` does not convert the versification, so it serves a different purpose. This does deviate from the C# implementation, but it has more consistent behavior.
___
*[`machine/scripture/verse_ref.py` line 488 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8LuK1HCjoOcCCDPH:-O4GrgJx19fZILcnMN4F:b-f4vr68) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L488)):*

I think there's actually a bug in SIL.Scripture/VerseRef.cs. The corresponding section in OverlappingVersesRanges for the VerseRef type input is `if (verseRef1.IsDefault || verseRef1.IsDefault)` but it should be `if (verseRef1.IsDefault || verseRef2.IsDefault)`.
</blockquote></details>

Yeah, that looks like a bug in the C# code.
___
*[`tests/corpora/test_scripture_ref.py` line 6 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GyDTh3zC1G2XNpdbE:-O4GyDTh3zC1G2XNpdbF:b-352bhf) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/tests/corpora/test_scripture_ref.py#L6)):*
> ```Python
> 
> 
> class TestScriptureRef(unittest.TestCase):
> ```

Is there some reason why you used a test case here? I would prefer it to be consistent with other test fixtures. If it is because you need to specify a description for each assert, you can do so with a normal [assert](https://doc.pytest.org/en/latest/how-to/assert.html#asserting-with-the-assert-statement).

mshannon-sil:
*[`machine/corpora/scripture_text.py` line 82 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4G_gqn-DCe8y5VGXrR:-O4HDmWFDnwgX8BSPrIX:blwd8um) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text.py#L82)):*

Because Python doesn't support overloading, we sometimes have to deviate from the C# code a bit, as you discovered for the `_create_rows` method. I think it would be good to name this something else, maybe `_create_scripture_row`, then you wouldn't have to call the base class `_create_row` method using `super()`. If you rename this method, then you should also rename the `_create_rows` method to match.
</blockquote></details>

Should I do this for the get_rows() method as well? Renaming it to get_scripture_rows() so that in the method, rather than calling super().get_rows(), I just call self.get_rows()?

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111)* status: all files reviewed, 8 unresolved discussions (waiting on @ddaspit)

___
*[`machine/scripture/scripture_element.py` line 10 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GnqNM0JtJke216huW:-O4HK6iwA6l5WibQeX_t:b-kyif2u) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_element.py#L10)):*

This class should be moved to the `corpora` package.  This class should also be exported from `__init__.py` in the `corpora` package.
</blockquote></details>

I realize I didn't export most of the classes, I'll go ahead and do that.
___
*[`machine/scripture/scripture_ref.py` line 13 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4Gla2142SMa4u_R7TO:-O4HIOlYBnLWrDfVAlE2:b-ov03ae) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L13)):*

This class should be moved to the `corpora` package.
</blockquote></details>

Okay. I had put it in the `scripture` folder since it seemed to parallel the VerseRef class in some ways. Is the idea of the `scripture` folder just to include python equivalents of `SIL.Scripture`?
___
*[`machine/scripture/verse_ref.py` line 488 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8LuK1HCjoOcCCDPH:-O4HLObGCHaRFlqDPbMq:b-m9i4gg) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L488)):*

Yeah, that looks like a bug in the C# code.
</blockquote></details>

Do we need to submit a PR to SIL.Scripture to fix it?
___
*[`tests/corpora/test_scripture_ref.py` line 6 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GyDTh3zC1G2XNpdbE:-O4HLs4HDihiAEjVhRe1:b60pwe) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/tests/corpora/test_scripture_ref.py#L6)):*

Is there some reason why you used a test case here? I would prefer it to be consistent with other test fixtures. If it is because you need to specify a description for each assert, you can do so with a normal [assert](https://doc.pytest.org/en/latest/how-to/assert.html#asserting-with-the-assert-statement).
</blockquote></details>

I was trying to follow the format of the corresponding test file in Machine, since it uses test cases unlike the other corpora test files. But I agree, it makes sense to keep it consistent. I'll go ahead and change it to match the other files in machine.py.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111)* status: all files reviewed, 8 unresolved discussions (waiting on @mshannon-sil)

___
*[`machine/corpora/scripture_text.py` line 82 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4G_gqn-DCe8y5VGXrR:-O4Hk8ME5kKEzvPWVF95:b-km0ujf) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text.py#L82)):*

Should I do this for the get_rows() method as well? Renaming it to get_scripture_rows() so that in the method, rather than calling super().get_rows(), I just call self.get_rows()?
</blockquote></details>

In the case of `get_rows`, we are overriding the method, so we don't want to change the name. It is also pretty normal to call the base class method when overriding, because you are often adding behavior to the base class implementation.
___
*[`machine/scripture/scripture_ref.py` line 13 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4Gla2142SMa4u_R7TO:-O4Hm_XH3YC9A8X6fMF2:bdkbqhj) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L13)):*

Okay. I had put it in the `scripture` folder since it seemed to parallel the VerseRef class in some ways. Is the idea of the `scripture` folder just to include python equivalents of `SIL.Scripture`?
</blockquote></details>

Yes, that is correct.
___
*[`machine/scripture/verse_ref.py` line 488 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8LuK1HCjoOcCCDPH:-O4HmfIhCqLCFWJx_yMT:b63wln8) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L488)):*

Do we need to submit a PR to SIL.Scripture to fix it?
</blockquote></details>

It would be the neighborly thing to do, but probably not a high priority.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111)* status: all files reviewed, 6 unresolved discussions (waiting on @ddaspit)

___
*[`machine/corpora/scripture_text.py` line 82 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4G_gqn-DCe8y5VGXrR:-O4LKepaCKzVt1FGqfGW:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text.py#L82)):*

In the case of `get_rows`, we are overriding the method, so we don't want to change the name. It is also pretty normal to call the base class method when overriding, because you are often adding behavior to the base class implementation.
</blockquote></details>

Done.
___
*[`machine/corpora/scripture_text_corpus.py` line 111 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GeBLIB3IMr3nHrkon:-O4LKggX9EIcfSKr8K8M:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/scripture_text_corpus.py#L111)):*

We want to export this function from the `__init__.py` of the `corpora` package.
</blockquote></details>

Done.
___
*[`machine/corpora/standard_parallel_text_corpus.py` line 255 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GbzFM3V9bl-0Fo0vW:-O4LKhw7ETQ1tPXfcwv0:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/corpora/standard_parallel_text_corpus.py#L255)):*

This was a bug in the C# code, you should use the `is_scripture` function.
</blockquote></details>

Done.
___
*[`machine/scripture/scripture_element.py` line 10 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GnqNM0JtJke216huW:-O4LKkB-2mwIwjPvJuAG:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_element.py#L10)):*

I realize I didn't export most of the classes, I'll go ahead and do that.
</blockquote></details>

Done.
___
*[`machine/scripture/scripture_ref.py` line 13 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4Gla2142SMa4u_R7TO:-O4LKpTUCV2EMTxce-di:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L13)):*

Yes, that is correct.
</blockquote></details>

Done.
___
*[`machine/scripture/scripture_ref.py` line 18 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C1XrA52lfXCyPQQgy:-O4LRnHL6y1t8Wen5NUl:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/scripture_ref.py#L18)):*

You can make this a top-level constant, i.e. `EMPTY_SCRIPTURE_REF`.
</blockquote></details>

Done.
___
*[`machine/scripture/verse_ref.py` line 380 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4C8ygt7zsh7lzn1xtp:-O4LX6u929ABzyeK-pZy:bmhtrwi) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/machine/scripture/verse_ref.py#L380)):*

The `Comparable` base class provides an implementation of `__eq__` based on the `compare_to` method. This makes all of the comparison methods consistent with each other. `exact_equals` does not convert the versification, so it serves a different purpose. This does deviate from the C# implementation, but it has more consistent behavior.
</blockquote></details>

Okay, I changed it back.
___
*[`tests/corpora/test_scripture_ref.py` line 6 at r6](https://reviewable.io/reviews/sillsdev/machine.py/111#-O4GyDTh3zC1G2XNpdbE:-O4Lhr4M50-MrKzaRPpj:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/fb630bf98952e872d2a0632ca0e6f56ac581f53f/tests/corpora/test_scripture_ref.py#L6)):*

I was trying to follow the format of the corresponding test file in Machine, since it uses test cases unlike the other corpora test files. But I agree, it makes sense to keep it consistent. I'll go ahead and change it to match the other files in machine.py.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> Reviewed 19 of 19 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/111#-:-O4MMuwY46gAe7Mbdv5i:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #141, opened 2024-11-11
Raise error when id tag doesn't match filename book id

mshannon-sil:
This PR addresses https://github.com/sillsdev/silnlp/issues/574. ParatextTextCorpus and ParatextBackupTextCorpus now raise a ValueError if the book id in the filename and the \id tag inside the file don't match for a given SFM file in the Paratext project. For example, if the filename is 07JDG.SFM but the \id tag is JUD, this will now raise an error during initialization, whereas before initialization would succeed without any message to the user and would use the incorrect \id tag for that book's verse refs. I added two error messages, one for if the \id tag itself is invalid, and another for if the \id tag is valid but does not match the book id in the filename.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/141)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/141)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @mshannon-sil)

___
*[`machine/corpora/paratext_backup_text_corpus.py` line 33 at r1](https://reviewable.io/reviews/sillsdev/machine.py/141#-OBWA6vt093BCRQfn_F7:-OBWA6vt093BCRQfn_F8:bhl1wzb) ([raw file](https://github.com/sillsdev/machine.py/blob/8679b785cdecb2a427726baf55b023a80496498e/machine/corpora/paratext_backup_text_corpus.py#L33)):*
> ```Python
>                         settings.name,
>                     )
>                     with text.get_rows() as rows:
> ```

We purposefully avoid parsing the book in the constructor. We want to avoid parsing errors in books that we filter out when actually iterating over the corpus. Can we perform this check when iterating over the corpus, i.e. when `get_rows` is called?

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/141)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @mshannon-sil)

___
*[`machine/corpora/paratext_backup_text_corpus.py` line 33 at r1](https://reviewable.io/reviews/sillsdev/machine.py/141#-OBWA6vt093BCRQfn_F7:-OBWsOlf3VJm-5coZI1A:b9qy24s) ([raw file](https://github.com/sillsdev/machine.py/blob/8679b785cdecb2a427726baf55b023a80496498e/machine/corpora/paratext_backup_text_corpus.py#L33)):*

We purposefully avoid parsing the book in the constructor. We want to avoid parsing errors in books that we filter out when actually iterating over the corpus. Can we perform this check when iterating over the corpus, i.e. when `get_rows` is called?
</blockquote></details>

That's fair, and yes I took a look and we should be able to do a similar check in `get_rows`, comparing the ref book to the text_id for each row. That means the check would happen for each row rather than once per SFM file, but it sounds like that's worth it to avoid parsing books in the constructor that we would have filtered out.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/141)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @mshannon-sil)

___
*[`machine/corpora/paratext_backup_text_corpus.py` line 33 at r1](https://reviewable.io/reviews/sillsdev/machine.py/141#-OBWA6vt093BCRQfn_F7:-OBXKqo60X7R-zqCY898:bw60k3a) ([raw file](https://github.com/sillsdev/machine.py/blob/8679b785cdecb2a427726baf55b023a80496498e/machine/corpora/paratext_backup_text_corpus.py#L33)):*

That's fair, and yes I took a look and we should be able to do a similar check in `get_rows`, comparing the ref book to the text_id for each row. That means the check would happen for each row rather than once per SFM file, but it sounds like that's worth it to avoid parsing books in the constructor that we would have filtered out.
</blockquote></details>

It would be ideal if we could just perform the check when we hit the `\id` marker.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/141)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @mshannon-sil)

___
*[`machine/corpora/paratext_backup_text_corpus.py` line 33 at r1](https://reviewable.io/reviews/sillsdev/machine.py/141#-OBWA6vt093BCRQfn_F7:-OBg_KkU2QPl2s_LkBvg:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/8679b785cdecb2a427726baf55b023a80496498e/machine/corpora/paratext_backup_text_corpus.py#L33)):*

It would be ideal if we could just perform the check when we hit the `\id` marker.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> Reviewed 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/141#-:-OBgpVNT6ZjBG88C_SLG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #161, opened 2025-03-06
Updates to USFM updater

johnml1135:
Add tests for beginning-of-verse embeds
Fix the embed at beginning issue (https://github.com/sillsdev/serval/issues/636)
Add paragraph marker control (https://github.com/sillsdev/serval/issues/627)
Always preserve \r and \rem
Correct behavior for stripping text behavior (https://github.com/sillsdev/serval/issues/629)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/161)

### DIFF
--- a/machine/corpora/paratext_project_text_updater_base.py
+++ b/machine/corpora/paratext_project_text_updater_base.py
@@ -22,6 +22,7 @@ class ParatextProjectTextUpdaterBase(ABC):
         rows: Optional[Sequence[Tuple[Sequence[ScriptureRef], str]]] = None,
         full_name: Optional[str] = None,
         text_behavior: UpdateUsfmTextBehavior = UpdateUsfmTextBehavior.PREFER_EXISTING,
+        paragraph_behavior: UpdateUsfmMarkerBehavior = UpdateUsfmMarkerBehavior.PRESERVE,
         embed_behavior: UpdateUsfmMarkerBehavior = UpdateUsfmMarkerBehavior.PRESERVE,
         style_behavior: UpdateUsfmMarkerBehavior = UpdateUsfmMarkerBehavior.STRIP,
     ) -> Optional[str]:
@@ -31,7 +32,12 @@ class ParatextProjectTextUpdaterBase(ABC):
         with self._open(file_name) as sfm_file:
             usfm: str = sfm_file.read().decode(self._settings.encoding)
         handler = UpdateUsfmParserHandler(
-            rows, None if full_name is None else f"- {full_name}", text_behavior, embed_behavior, style_behavior
+            rows,
+            None if full_name is None else f"- {full_name}",
+            text_behavior,
+            paragraph_behavior,
+            embed_behavior,
+            style_behavior,
         )
         try:
             parse_usfm(usfm, handler, self._settings.stylesheet, self._settings.versification)
--- a/machine/corpora/paratext_project_text_updater_base.py
+++ b/machine/corpora/paratext_project_text_updater_base.py
@@ -25,6 +25,7 @@ class ParatextProjectTextUpdaterBase(ABC):
         paragraph_behavior: UpdateUsfmMarkerBehavior = UpdateUsfmMarkerBehavior.PRESERVE,
         embed_behavior: UpdateUsfmMarkerBehavior = UpdateUsfmMarkerBehavior.PRESERVE,
         style_behavior: UpdateUsfmMarkerBehavior = UpdateUsfmMarkerBehavior.STRIP,
+        preserve_paragraph_styles: Optional[Sequence[str]] = None,
     ) -> Optional[str]:
         file_name: str = self._settings.get_book_file_name(book_id)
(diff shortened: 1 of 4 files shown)


ddaspit (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/161)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`machine/corpora/update_usfm_parser_handler.py` line 360 at r1](https://reviewable.io/reviews/sillsdev/machine.py/161#-OKhEUxq5faeyC2SwgV4:-OKhEUxq5faeyC2SwgV5:bz0ul1z) ([raw file](https://github.com/sillsdev/machine.py/blob/ec6d78b09246db9ba0e8e543f01c802090c2d76b/machine/corpora/update_usfm_parser_handler.py#L360)):*
> ```Python
> 
>     def _has_new_text(self) -> bool:
>         return bool(self._replace_stack) and self._replace_stack[-1]
> ```

Is this testing if `_replace_stack` is empty? If so, I would make the intent more explicit by checking the length.

isaac091:
When updating a \r or \rem, the old text is kept in.

Original:
> \id MAT - Test
> \c 1
> \v 1 Hello World
> \r reference

Rows: `[(MAT 1:1/1:r, "new reference text")]`

Output: (default settings on updater)
> \id MAT - Test
> \c 1
> \v 1 Hello World
> \r new reference text reference

isaac091:
Another potential inconsistency is the interaction between `paragraph_behavior=STRIP` and the "paragraph" markers that function like embeds (\r, \rem, \s1, etc.).

Original:
> \id MAT - Test
> \c 1
> \v 1 Hello World
> \s1 section header

Rows: `[(MAT 1:1, "new verse 1")]`

Output: (default settings besides `paragraph_behavior=STRIP`)
> \id MAT - Test
> \c 1
> \v 1 new verse 1 section header

Generally, these markers are treated as separate from the verse before them, despite the fact that they share a VerseRef. In this case, they're being treated half like they're separate from the verse (content is kept whereas the text after a \p would not be) and half like they're a part of the verse (marker is deleted and content is put in the main verse line). 

I tested this on the initial paragraph behavior commit and it was happening there as well.

ddaspit (commented):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/161)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

johnml1135:
*[`machine/corpora/update_usfm_parser_handler.py` line 360 at r1](https://reviewable.io/reviews/sillsdev/machine.py/161#-OKhEUxq5faeyC2SwgV4:-OKm453z84u53HkwJNzS:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/ec6d78b09246db9ba0e8e543f01c802090c2d76b/machine/corpora/update_usfm_parser_handler.py#L360)):*

Is this testing if `_replace_stack` is empty? If so, I would make the intent more explicit by checking the length.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/161#-:-OKm7x2W76hnp_5n8TCw:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

johnml1135:
How should we handle these two cases?  What do you think @ddaspit?  I am unsure:
* For reference update/not update, this should never happeen in Serval because it will not send these to be pretranslated.  Should we override if there is something there or just ignore it?
* For Section headers, should we create a new category where it becomes "non-verse"?  I need to dig into it more.  We really only want to strip paragraph markers if they are in a verse proper.  What logic can we use to do that?

ddaspit (commented):
- We should override if there is something there. Now that I think about it a bit more, maybe we should handle `\r` and `\rem` markers in Serval instead of in Machine. The text updater class is supposed to be a general-purpose class for replacing text in USFM. Could we pass through `\r` and `\rem` marker text from the source without translating it?
- We should only strip paragraph markers that occur in the middle of a verse. Section headers will never occur in a verse. You should be able to use `CurrentTextType` to determine if you are in a verse.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/161#-:-OKmFCjt6NdwIIKQaG2w:b-nah73y)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

Enkidu93 (commented):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/161)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

PR #179, opened 2025-05-19
Fixes for generating alignments alongside pretranslations

Enkidu93:
* Make `align_pretranslations` configurable in `build_options`
* Don't try to ingest non-existent `pretranslation.src.json` properties

Related to https://github.com/sillsdev/serval/issues/663


Enkidu93:
While I was digging into this testing it and trying to get some sample output pretranslation jsons, I realized that it's not particularly straightforward to run machine.py nmt jobs locally end-to-end. Is there an issue already for making a dummy translation job for testing? What's in the launch.json doesn't work well locally and the code isn't set up for using the cpu.

PR #195, opened 2025-06-16
Align pretranslations by default

Enkidu93:
Change requested [here](https://github.com/sillsdev/serval/pull/703#pullrequestreview-2915053074).

### DIFF
--- a/machine/jobs/build_nmt_engine.py
+++ b/machine/jobs/build_nmt_engine.py
@@ -55,8 +55,6 @@ def run(args: dict) -> None:
             except TypeError as e:
                 raise TypeError(f"Build options could not be parsed: {e}") from e
             SETTINGS.update({model_type: build_options})
-            if "align_pretranslations" in build_options:
-                SETTINGS.update({"align_pretranslations": build_options["align_pretranslations"]})
         SETTINGS.data_dir = os.path.expanduser(cast(str, SETTINGS.data_dir))
 
         logger.info(f"Config: {SETTINGS.as_dict()}")
--- a/machine/jobs/settings.yaml
+++ b/machine/jobs/settings.yaml
@@ -3,7 +3,7 @@ default:
   shared_file_uri: s3:/silnlp/
   shared_file_folder: production
   inference_batch_size: 1024
-  align_pretranslations: false
+  align_pretranslations: true
   huggingface:
     parent_model_name: facebook/nllb-200-distilled-1.3B
     train_params:


Enkidu93:
*[`machine/jobs/build_nmt_engine.py` line 58 at r1](https://reviewable.io/reviews/sillsdev/machine.py/195#-OT7e_RP5O54tyNzJTqx:-OT7uu9Z3n7RcdxoAVNy:b-opzro6) ([raw file](https://github.com/sillsdev/machine.py/blob/a6d5a1f0a2a36f9a09e361b68b52d5a4a5ee8b2b/machine/jobs/build_nmt_engine.py#L58)):*

I don't think we want to remove this. We want a way to disable alignments using build options.
</blockquote></details>

Do you expect that we'd need to be able to turn this off from the client side though? My thinking was that now SF or some other client could turn off alignments at build time and then when they go to fetch USFM and place markers, it won't work properly, but there won't be visibility as to why. As it stands, this is something we could still turn off from within Serval when we define the machine.py script I believe.

Enkidu93 (commented):
Oh, yes :laughing:, thank you. I'm so used to :lgtm: = approved. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/195#-:-OT7ue-W5GjoIo25ER9v:b-1sgb14)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/195)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`machine/jobs/build_nmt_engine.py` line 58 at r1](https://reviewable.io/reviews/sillsdev/machine.py/195#-OT7e_RP5O54tyNzJTqx:-OT8shODBMNuVBIe8MUU:b-8qs7qg) ([raw file](https://github.com/sillsdev/machine.py/blob/a6d5a1f0a2a36f9a09e361b68b52d5a4a5ee8b2b/machine/jobs/build_nmt_engine.py#L58)):*

Do you expect that we'd need to be able to turn this off from the client side though? My thinking was that now SF or some other client could turn off alignments at build time and then when they go to fetch USFM and place markers, it won't work properly, but there won't be visibility as to why. As it stands, this is something we could still turn off from within Serval when we define the machine.py script I believe.
</blockquote></details>

I was thinking it might be useful to turn it off for testing purposes.

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/195#-:-OT98K3v9f951CHL2v_C:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Issue #193, opened 2025-06-10
Refactor `UpdateUsfmParserHandler` to accept rows with custom metadata


ddaspit:
Currently, the `UpdateUsfmParserHandler` class models a row as a tuple of scripture refs and a translation. Update block handlers might need additional metadata in order to process an update block properly. For example, the `PlaceMarkersUsfmUpdateBlockHandler` class needs alignment info for a row. Update block handlers have to manage and lookup this metadata itself. It would be better if we could pass in additional metadata for a row to the `UpdateUsfmParserHandler`. The metadata could then be included in an update block that is passed to update block handlers.

PR #202, opened 2025-06-26
Pass metadata through update block

Enkidu93:
Fixes https://github.com/sillsdev/machine.py/issues/193

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/202)

### DIFF
--- a/machine/corpora/__init__.py
+++ b/machine/corpora/__init__.py
@@ -51,7 +51,12 @@ from .token_processors import (
     normalize,
     unescape_spaces,
 )
-from .update_usfm_parser_handler import UpdateUsfmMarkerBehavior, UpdateUsfmParserHandler, UpdateUsfmTextBehavior
+from .update_usfm_parser_handler import (
+    UpdateUsfmMarkerBehavior,
+    UpdateUsfmParserHandler,
+    UpdateUsfmTextBehavior,
+    UpdateUsfmRow,
+)
 from .usfm_file_text import UsfmFileText
 from .usfm_file_text_corpus import UsfmFileTextCorpus
 from .usfm_memory_text import UsfmMemoryText
@@ -135,6 +140,7 @@ __all__ = [
     "UpdateUsfmMarkerBehavior",
     "UpdateUsfmParserHandler",
     "UpdateUsfmTextBehavior",
+    "UpdateUsfmRow",
     "UsfmAttribute",
     "UsfmElementType",
     "UsfmFileText",
--- a/machine/corpora/__init__.py
+++ b/machine/corpora/__init__.py
@@ -54,8 +54,8 @@ from .token_processors import (
 from .update_usfm_parser_handler import (
     UpdateUsfmMarkerBehavior,
     UpdateUsfmParserHandler,
-    UpdateUsfmTextBehavior,
     UpdateUsfmRow,
+    UpdateUsfmTextBehavior,
 )
 from .usfm_file_text import UsfmFileText
 from .usfm_file_text_corpus import UsfmFileTextCorpus
--- a/machine/corpora/paratext_project_text_updater_base.py
+++ b/machine/corpora/paratext_project_text_updater_base.py
@@ -5,7 +5,12 @@ from ..utils.typeshed import StrPath
 from .paratext_project_settings import ParatextProjectSettings
 from .paratext_project_settings_parser_base import ParatextProjectSettingsParserBase
 from .scripture_ref import ScriptureRef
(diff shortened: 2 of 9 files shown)


ddaspit (changes requested):
Yes, I think it makes sense to define `alignment_info` as a constant.

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/202#-:-OU1hBd-B1-44gP6qTFB:b-rnfp9c)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

___
*[`machine/corpora/place_markers_usfm_update_block_handler.py` line 12 at r1](https://reviewable.io/reviews/sillsdev/machine.py/202#-OU1gbx8Aqqv8ozZJPbz:-OU1gbx8Aqqv8ozZJPc-:beujpxb) ([raw file](https://github.com/sillsdev/machine.py/blob/cf07e8ec877de68eaca43873c6c168c295df4eda/machine/corpora/place_markers_usfm_update_block_handler.py#L12)):*
> ```Python
> 
> 
> class PlaceMarkersAlignmentInfo(TypedDict):
> ```

I think this is only used in unit tests. We could cast `alignment_info` to this class when we use it. See my comment below. If we don't use it here, we should just move it to the unit test.
___
*[`machine/corpora/place_markers_usfm_update_block_handler.py` line 28 at r1](https://reviewable.io/reviews/sillsdev/machine.py/202#-OU1eApo4akdrKlSnEUT:-OU1eApo4akdrKlSnEUU:b-4wdsyw) ([raw file](https://github.com/sillsdev/machine.py/blob/cf07e8ec877de68eaca43873c6c168c295df4eda/machine/corpora/place_markers_usfm_update_block_handler.py#L28)):*
> ```Python
>             len(elements) == 0
>             or "alignment_info" not in block.metadata
>             or not isinstance(block.metadata["alignment_info"], dict)
> ```

It is correct to thoroughly validate the metadata before using it. We could probably get away with doing less. You could just check that `alignment_info` exists, and then cast it to `PlaceMarkersAlignmentInfo`. We just assume that the data structure is correct if `alignment_info` exists. If it isn't correct, it will probablly cause a crash, which is fine, because it would indicate a coding error.

Enkidu93 (commented):
OK, done.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/202#-:-OU5Fc1E5mzh230OXyjX:bm3yqt3)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @isaac091)

___
*[`machine/corpora/place_markers_usfm_update_block_handler.py` line 12 at r1](https://reviewable.io/reviews/sillsdev/machine.py/202#-OU1gbx8Aqqv8ozZJPbz:-OU5FeMF2Ol-CpPGJHZG:b1ubcge) ([raw file](https://github.com/sillsdev/machine.py/blob/cf07e8ec877de68eaca43873c6c168c295df4eda/machine/corpora/place_markers_usfm_update_block_handler.py#L12)):*

I think this is only used in unit tests. We could cast `alignment_info` to this class when we use it. See my comment below. If we don't use it here, we should just move it to the unit test.
</blockquote></details>

Makes sense
___
*[`machine/corpora/place_markers_usfm_update_block_handler.py` line 28 at r1](https://reviewable.io/reviews/sillsdev/machine.py/202#-OU1eApo4akdrKlSnEUT:-OU5FmRI32rPDzOT0RSR:bm3yqt3) ([raw file](https://github.com/sillsdev/machine.py/blob/cf07e8ec877de68eaca43873c6c168c295df4eda/machine/corpora/place_markers_usfm_update_block_handler.py#L28)):*

It is correct to thoroughly validate the metadata before using it. We could probably get away with doing less. You could just check that `alignment_info` exists, and then cast it to `PlaceMarkersAlignmentInfo`. We just assume that the data structure is correct if `alignment_info` exists. If it isn't correct, it will probablly cause a crash, which is fine, because it would indicate a coding error.
</blockquote></details>

OK, done.
___
*[`machine/corpora/update_usfm_parser_handler.py` line 314 at r1](https://reviewable.io/reviews/sillsdev/machine.py/202#-OTi0F8b669cE9ldoGJ9:-OU5FsERAPV24QqAbsXp:bb6u032) ([raw file](https://github.com/sillsdev/machine.py/blob/cf07e8ec877de68eaca43873c6c168c295df4eda/machine/corpora/update_usfm_parser_handler.py#L314)):*
<details><summary><i>Previously, isaac091 (Isaac Schifferer) wrote…</i></summary><blockquote>

No issues that I can think of
</blockquote></details>

Cool

PR #203, opened 2025-06-28
Quotation denormalization

benjaminking:
I am creating a PR for quotation denormalization even though it still needs to be rebased.  Eli had thought it would be helpful to go through it and make some comments, even if there were some minor changes still coming.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)


benjaminking:
> Unused function

Removed


> Would `_have_depth_and_direction...` be better?

I have changed this, along with several other similarly-named functions.

> Why not also check the other properties?

In this case, the object has references to previous and next instances (a linked list basically), so including it in the equality function would lead to infinite recursion.

> I don't think many of these helper methods are necessary. In cases where the field is not read-only and there's no other logic in the getters and setters, why not just use the bare variable - i.e., `text_segment.previous_segment` rather than `set_...`/`get_...`? If it is read-only, you can use the `@property` decorator (you can search for lots of examples in the machine.py codebase). That would cut down the size of this boiler-plate code a lot. It would also make the logic elsewhere more succinct and readable.

I have made these changes throughout my code.  Thanks for the suggestion.

> Unused function

Removed.

> What about `substring_before(start_index) + replacement + substring_after(start_index)`?

Good suggestion.  Done.

> Maybe `next_character_matches()`? I think `<noun>_<verb>(e)s()` is cleaner than `does_<noun>_<verb>()`. It's shorter, it sounds more natural in `if` statements, and I don't think `does_...` is used elsewhere in machine.py.

Changed as part of the naming changes I mentioned earlier.

I also did a fair bit of minor refactoring to make things more consistent (that will be the majority of the changes in the commit).

Enkidu93 (changes requested):
Great!

> Reviewed 1 of 39 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OU6Qp1S-cAJEC6pWf74:btojbfd)* status: 2 of 49 files reviewed, 1 unresolved discussion

___
*[`machine/corpora/analysis/text_segment.py` line 25 at r2](https://reviewable.io/reviews/sillsdev/machine.py/203#-OU6Ri291k9ioBNNH3aO:-OU6Ri291k9ioBNNH3aP:b-dpw3sg) ([raw file](https://github.com/sillsdev/machine.py/blob/550d8a9521fe9c9bbb42cc428e9964f2d2ab6fc6/machine/corpora/analysis/text_segment.py#L25)):*
> ```Python
>         if self._index_in_verse != value._index_in_verse:
>             return False
>         if self._index_in_verse != value._index_in_verse:
> ```

I don't think you meant to check this twice. Maybe you meant to check one or more of the other non-`TextSegment` properties?

Enkidu93:
I'll try and give it a more thorough review soon - thank you for the refactor and cleanup! I'll also begin porting more of the code now. I'm sure Damien will have more comments :zany_face:.

Enkidu93 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)* status: 2 of 49 files reviewed, 4 unresolved discussions (waiting on @benjaminking)

___
*[`machine/corpora/analysis/quote_convention.py` line 73 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBaZsW5SrJRLnoItVZ:-OUBaZsW5SrJRLnoItV_:b-qmk1x7) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention.py#L73)):*
> ```Python
> 
>     def get_expected_quotation_mark(self, depth: int, direction: QuotationMarkDirection) -> str:
>         if depth > len(self.levels) or depth < 1:
> ```

Could use `get_num_levels`/`num_levels` 
___
*[`machine/corpora/analysis/quote_convention.py` line 122 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBairZ8rWp7RESO6Ei:-OUBairZ8rWp7RESO6Ej:b-xxeqhs) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention.py#L122)):*
> ```Python
>         return QuoteConvention(self.get_name() + "_normalized", [level.normalize() for level in self.levels])
> 
>     def print_summary(self) -> None:
> ```

Damien might have other thoughts. But I don't like having printing methods in non-command-line-y classes. Maybe instead of `_get_summary_message`, you could override `__repr__` or `__str__`.

ddaspit (changes requested):
> Reviewed 10 of 49 files at r1, 37 of 39 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)* status: all files reviewed, 16 unresolved discussions (waiting on @benjaminking and @Enkidu93)

___
*[`machine/corpora/analysis/quote_convention.py` line 122 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBairZ8rWp7RESO6Ei:-OUBelTt7kjHC8ZvP5Jw:btd51a6) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention.py#L122)):*

Damien might have other thoughts. But I don't like having printing methods in non-command-line-y classes. Maybe instead of `_get_summary_message`, you could override `__repr__` or `__str__`.
</blockquote></details>

I agree.
___
*[`machine/corpora/analysis/quotation_mark_tabulator.py` line 76 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUCA6qd9aw7h2p9uafT:-OUCA6qd9aw7h2p9uafU:bulammo) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_tabulator.py#L76)):*
> ```Python
>         return 1 - (num_differences / num_total_quotation_marks)
> 
>     def print_summary(self) -> None:
> ```

Consider using `__str__` or `__rep__`. If that is inappropriate, you can have the method return a string description instead of calling `print` directly.
___
*[`machine/corpora/analysis/quotation_mark_finder.py` line 13 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC9NbIEIqScrH2ebrg:-OUC9NbIEIqScrH2ebrh:b2fcuml) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_finder.py#L13)):*
> ```Python
> 
> class QuotationMarkFinder:
>     quote_pattern = regex.compile(r"(\p{Quotation_Mark}|<<|>>|<|>)", regex.U)
> ```

This should be all caps and probably prefixed with an underscore.
___
*[`machine/corpora/scripture_embed.py` line 0 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBgdMz5_7z_Mz7Xi7o:-OUBgdMz5_7z_Mz7Xi7p:bmlslt0) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/scripture_embed.py#L0)):*
Is this module used anywhere?
___
*[`machine/corpora/analysis/quotation_mark_resolution_settings.py` line 28 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC8iU08rI9qdqrmHRq:-OUC8iU08rI9qdqrmHRr:bn9kv4p) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_resolution_settings.py#L28)):*
> ```Python
> 
>     @abstractmethod
>     def should_rely_on_paragraph_markers(self) -> bool: ...
> ```

This could be a property.
___
*[`machine/corpora/analysis/quote_convention_set.py` line 161 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC7x7E0OPkTb9875T1:-OUC7x7E0OPkTb9875T2:bb3t41p) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L161)):*
> ```Python
>         return (best_quote_convention, best_similarity)
> 
>     def print_summary(self) -> None:
> ```

Consider using `__str__` or `__rep__` instead of direct `print` calls.
___
*[`machine/corpora/analysis/quote_convention_detector.py` line 18 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC8J1cEHrg6MCribqm:-OUC8J1cEHrg6MCribqn:bv2xub) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_detector.py#L18)):*
> ```Python
> 
> 
> @dataclass
> ```

Consider marking this dataclass as immutable.
___
*[`machine/corpora/analysis/standard_quote_conventions.py` line 4 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC7kE--085AAxJAbLY:-OUC7kE--085AAxJAbLZ:b-9b3oj9) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/standard_quote_conventions.py#L4)):*
> ```Python
> from .quote_convention_set import QuoteConventionSet
> 
> standard_quote_conventions: QuoteConventionSet = QuoteConventionSet(
> ```

The naming convention for constants is all caps.
___
*[`machine/corpora/analysis/quotation_mark_string_match.py` line 16 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC99hv0hEfH_VmHFE6:-OUC99hv0hEfH_VmHFE7:b-2uwisv) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_string_match.py#L16)):*
> ```Python
> 
>     # extra stuff in the regex to handle Western Cham
>     letter_pattern: Pattern = regex.compile(r"[\p{L}\U0001E200-\U0001E28F]", regex.U)
> ```

You should use all caps for constant names. I think this are only used internally, so they should be prefixed with an underscore.
___
*[`machine/corpora/analysis/quote_convention.py` line 6 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC6Wa80BYihTM2g8pc:-OUC6Wa80BYihTM2g8pd:bhyf3jm) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention.py#L6)):*
> ```Python
> from .quotation_mark_direction import QuotationMarkDirection
> 
> quote_normalization_map: Dict[str, str] = {
> ```

The naming convention for constants is all caps. If the constant is only used locally in the module, then you should prefix with an underscore.
___
*[`machine/corpora/analysis/quote_convention.py` line 22 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC7KQr7_zGV06rwSh6:-OUC7KQr7_zGV06rwSh7:bikzp89) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention.py#L22)):*
> ```Python
> 
> 
> @dataclass
> ```

If a dataclass is intended to be immutable, consider using `@dataclass(frozen=True)`.
___
*[`machine/corpora/analysis/__init__.py` line 0 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBg86kAgMgD1jd93Hr:-OUBg86kAgMgD1jd93Hs:bss12lc) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/__init__.py#L0)):*
I think this package name isn't specific enough, unless you were thinking that other kinds of "analysis" classes might end up here in the future. What about `punctuation_analysis`?
___
*[`machine/corpora/analysis/depth_based_quotation_mark_resolver.py` line 104 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUC9qaa9r0Ye_edIOOb:-OUC9qaa9r0Ye_edIOOc:bb9i9s8) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/depth_based_quotation_mark_resolver.py#L104)):*
> ```Python
> 
> class QuotationMarkCategorizer:
>     apostrophe_pattern = regex.compile(r"[\'\u2019\u2018]", regex.U)
> ```

This should be all caps and prefixed with an underscore.

Enkidu93 (changes requested):
Another batch of comments - wanted to get a chunk in so that they aren't all coming in at once. 

> Reviewed 1 of 39 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OUaHGJ75SZPc6Jiu0vJ:bny1rb6)* status: all files reviewed, 28 unresolved discussions (waiting on @benjaminking)

___
*[`machine/corpora/analysis/quote_convention_set.py` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaF0UOA-hTOcjkyQju:-OUaF0UOA-hTOcjkyQjv:b-bzbsgn) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L25)):*
> ```Python
>         opening_quotation_marks: Set[str] = set()
>         closing_quotation_marks: Set[str] = set()
>         all_quotation_marks: Set[str] = set()
> ```

Since this is just the union of the other two, could we just set it as such downstream?
___
*[`machine/corpora/analysis/quote_convention_set.py` line 27 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaF5Ul8JkJxhfH_jyM:-OUaF5Ul8JkJxhfH_jyN:b-d5ry5d) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L27)):*
> ```Python
>         all_quotation_marks: Set[str] = set()
> 
>         if len(self._conventions) > 0:
> ```

I don't think this if statement is needed
___
*[`machine/corpora/analysis/quote_convention_set.py` line 39 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaF9Ty5MGWna1RAj69:-OUaF9Ty5MGWna1RAj6A:b48tlgg) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L39)):*
> ```Python
>             if len(all_quotation_marks) > 0:
>                 self._opening_quotation_mark_regex: Pattern = regex.compile(
>                     r"[" + "".join(sorted(list(opening_quotation_marks))) + "]"
> ```

Why are these all sorted (and elsewhere there's similar sorting)? Is it mainly for testing or something like that?
___
*[`machine/corpora/analysis/quote_convention_set.py` line 49 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaFNkC2_7TWiEXIqa0:-OUaFNkC2_7TWiEXIqa1:b-l4a6z2) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L49)):*
> ```Python
> 
>         if len(opening_quotation_marks) == 0:
>             self._opening_quotation_mark_regex = regex.compile(r"")
> ```

I don't know how computationally expensive it would be to just make these the defaults and avoid these if statements? Or would that not work in some cases?
___
*[`machine/corpora/analysis/quote_convention_set.py` line 55 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaFi9TA77H9Ep2gFqe:-OUaFi9TA77H9Ep2gFqf:burh3gm) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L55)):*
> ```Python
>             self._all_quotation_mark_regex = regex.compile(r"")
> 
>     def _create_quotation_mark_pair_map(self) -> None:
> ```

Could this function be combined with the above function without the combined function getting too hairy? Since they are called in the same place as a way of initializing everything and are iterating through the same collection in the same way, it would save a few lines and possibly be cleaner (?) 
___
*[`machine/corpora/analysis/quote_convention_set.py` line 88 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaG9nR2Trjqs97LysJ:-OUaG9nR2Trjqs97LysK:byw7qa9) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L88)):*
> ```Python
> 
>     def get_all_quote_convention_names(self) -> List[str]:
>         return sorted([qc.name for qc in self._conventions])
> ```

More sorting - why is this needed?
___
*[`machine/corpora/analysis/quotation_mark_string_match.py` line 57 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaDEYI2-I5_LvWhvGv:-OUaDEYI2-I5_LvWhvGw:b-ieyh1c) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_string_match.py#L57)):*
> ```Python
>     @property
>     def previous_character(self) -> Union[str, None]:
>         if self._start_index == 0:
> ```

Use `is_at_start_of_segment`? 
___
*[`machine/corpora/analysis/quotation_mark_tabulator.py` line 8 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaDWso3XJk2_Kub5SE:-OUaDWso3XJk2_Kub5SF:b-zcy81j) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_tabulator.py#L8)):*
> ```Python
> 
> 
> class QuotationMarkCounts:
> ```

I wonder if you could leverage python's `Counter` class to simplify this or remove it altogether?
___
*[`machine/corpora/analysis/quotation_mark_tabulator.py` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaDg1N1x3IMPTRGtGT:-OUaDg1N1x3IMPTRGtGU:bkiki8o) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_tabulator.py#L10)):*
> ```Python
> class QuotationMarkCounts:
>     def __init__(self):
>         self._string_counts: Dict[str, int] = dict()
> ```

Could this name be more specific? 
___
*[`machine/corpora/analysis/quotation_mark_tabulator.py` line 35 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaDzjSBIZqPcWBCsD-:-OUaDzjSBIZqPcWBCsD0:bl9rml4) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_tabulator.py#L35)):*
> ```Python
> 
>     def __init__(self):
>         self.quotation_counts_by_depth_and_direction: dict[tuple[int, QuotationMarkDirection], QuotationMarkCounts] = (
> ```

Should this be made private, as it were? 
___
*[`machine/corpora/analysis/quotation_mark_tabulator.py` line 43 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaEFZI3buWRlz_mcsH:-OUaEFZI3buWRlz_mcsI:b-9d0mrs) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_tabulator.py#L43)):*
> ```Python
>             self._count_quotation_mark(quotation_mark)
> 
>     def _count_quotation_mark(self, quote: QuotationMarkMetadata) -> None:
> ```

Throughout the code, there seems to be inconsistent use of `quote`, `quotationMark`, and `quotation` in method names and variables (or at least I haven't figured out the pattern). It would be nice to make this consistent.
___
*[`machine/corpora/analysis/quotation_mark_tabulator.py` line 59 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaEgzl3kxALZK3gS6v:-OUaEgzl3kxALZK3gS6w:b47g4mh) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_tabulator.py#L59)):*
> ```Python
> 
>     def calculate_similarity(self, quote_convention: QuoteConvention) -> float:
>         num_differences = 0
> ```

Could we qualify this name `weighted_count...` or something? The name suggests to me that it's an int but it's really a float.

Enkidu93 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)* status: all files reviewed, 32 unresolved discussions (waiting on @benjaminking)

___
*[`machine/corpora/analysis/preliminary_quotation_analyzer.py` line 43 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUapswL2tPfqA7Jks7w:-OUapswM7wj0F0113gQ6:bvitlk3) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/preliminary_quotation_analyzer.py#L43)):*
> ```Python
> 
>     def count_word_initial_apostrophe(self, quotation_mark: str) -> None:
>         if quotation_mark not in self._word_initial_occurrences:
> ```

This very same pattern of checking if a key exists then updating the value based on the previous value happens a lot in the code. It might be worth using python's `defaultdict` to avoid this. I have an extension in C# `UpdateValue` that I can use, I think, to a similar end. 
___
*[`machine/corpora/analysis/preliminary_quotation_analyzer.py` line 76 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaqrQf2-1RMq5o9MTk:-OUaqrQf2-1RMq5o9MTl:b-tfglwx) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/preliminary_quotation_analyzer.py#L76)):*
> ```Python
>         num_initial_marks: int = self._get_word_initial_occurrences(quotation_mark)
>         num_total_marks: int = self._get_total_occurrences(quotation_mark)
>         return num_total_marks > 0 and num_initial_marks / num_total_marks < 0.1
> ```

This class uses a lot of "magic numbers". I don't know if it would be possible to describe them and put them into constants??
___
*[`machine/corpora/analysis/preliminary_quotation_analyzer.py` line 103 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUarJxy3KOnlTYFfYRC:-OUarJxy3KOnlTYFfYRD:btfjgfp) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/preliminary_quotation_analyzer.py#L103)):*
> ```Python
>         self._later_quotation_mark_counts: Dict[str, int] = dict()
> 
>     def record_earlier_quotation_mark(self, quotation_mark: str) -> None:
> ```

Above you use `count...()` for method names but here you use `record...()`.
___
*[`machine/corpora/analysis/preliminary_quotation_analyzer.py` line 155 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUarUfl193v2zuPWZqJ:-OUarUfl193v2zuPWZqK:b-mpcnz7) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/preliminary_quotation_analyzer.py#L155)):*
> ```Python
>         self._group_quotation_marks(quotation_marks)
> 
>     def _group_quotation_marks(self, quotation_marks: List[QuotationMarkStringMatch]) -> None:
> ```

You could use `functools.reduce` or `itertools.grouby` to make this a one-liner. It looks like we use `groupby` in `thot_smt_model_trainer.py`.

Enkidu93 (changes requested):
Thanks! Great. I'll look back through the comments and try to confirm that all have been addressed. If you could please reply in Reviewable within the thread specific to that comment, it would make the back and forth a lot easier as we continue. 

I went ahead and ported the changes you've made so far. 

This should be my last round of new comments :zany_face: :crossed_fingers:. 

> It looks like neither of these are used in the project so far. Does it seem worth adding another library for this? I was able to reduce this to three lines using `defaultdict`.
`itertools` is in the poetry lock and `groupby` is used in `thot_smt_model_trainer.py` - what do you mean by 'project'? It's not a big difference if it's left as-is, but it would simplify the function.

> I'm not sure I see why it's incorrect. Can you expound?
I see now. I misread it and thought it had been copy-pasted from the previous error message. I like the new error message you came up with better :thumbsup:. 

> Reviewed 1 of 49 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OV8iZU4DsS4UzOITDzA:bnfk1v5)* status: 3 of 49 files reviewed, 57 unresolved discussions (waiting on @benjaminking and @ddaspit)

___
*[`machine/corpora/analysis/text_segment.py` line 70 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUkp3mI456EkQHH5lF-:-OV8u7wS1N_ZqKdtF7pS:bp0vel0) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/text_segment.py#L70)):*

Since these properties can be get'd and set'd and there's no special logic on either end, why not just make them public?
</blockquote></details>

Since there's no custom logic in the getter or setter, I'd just ditch the properties and replace e.g. `_previous_segment` with `previous_segment`.
___
*[`machine/corpora/quotation_mark_update_first_pass.py` line 37 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8ltuT3mlMCh-C1sji:-OV8ltuT3mlMCh-C1sjj:bp8l3td) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quotation_mark_update_first_pass.py#L37)):*
> ```Python
>     ) -> bool:
>         target_marks_by_source_marks: Dict[str, Set[str]] = {}
>         for level in range(1, source_quote_convention.num_levels + 1):
> ```

Some places `depth` is used and elsewhere `level`. Is there a true difference in meaning?
___
*[`machine/corpora/quotation_mark_update_first_pass.py` line 71 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8mTLp8cNYH0LJwJyB:-OV8mTLp8cNYH0LJwJyC:b-g4rg7k) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quotation_mark_update_first_pass.py#L71)):*
> ```Python
>         return self._choose_best_strategy_based_on_observed_issues(self._quotation_mark_resolver.get_issues())
> 
>     def _choose_best_strategy_based_on_observed_issues(self, issues) -> QuotationMarkUpdateStrategy:
> ```

Type-hinting for `issues`
___
*[`machine/corpora/punctuation_analysis/text_segment.py` line 78 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8u3s-3zZJMDrYbc0L:-OV8u3s-3zZJMDrYbc0M:bycx6ef) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/punctuation_analysis/text_segment.py#L78)):*
> ```Python
>         self._next_segment = next_segment
> 
>     def set_index_in_verse(self, index_in_verse: int) -> None:
> ```

Same for `index_in_verse` as the segment fields.
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 69 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8o0fSDaZulR74AbHn:-OV8o0fSDaZulR74AbHo:b-vp6is9) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L69)):*
> ```Python
> 
>     def _apply_fallback_updating(self, block: UsfmUpdateBlock) -> UsfmUpdateBlock:
>         for element in block._elements:
> ```

Use `Elements`
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 74 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8o3w13vLEfyNPwv4p:-OV8o3w13vLEfyNPwv4q:b-vp6is9) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L74)):*
> ```Python
> 
>     def _apply_standard_updating(self, block: UsfmUpdateBlock) -> UsfmUpdateBlock:
>         for element in block._elements:
> ```

Use `Elements`
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 111 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8rD6aBesiNxy0MDB8:-OV8rD6aBesiNxy0MDB9:b7yx9vn) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L111)):*
> ```Python
> 
>     def _create_text_segment(self, token: UsfmToken) -> Union[TextSegment, None]:
>         self._next_scripture_text_segment_builder.set_usfm_token(token)
> ```

Might be a little cleaner to do something like:
```
text_segment_to_return = null
if token.text != null:
    ...
return text_segment_to_return
```
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 132 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8sDjG-PA6z_8zhgVt:-OV8sDjG-PA6z_8zhgVu:b-cw3bpn) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L132)):*
> ```Python
>             if scripture_ref.chapter_num != self._current_chapter_number:
>                 self._current_chapter_number = scripture_ref.chapter_num
>                 self._start_new_chapter(self._current_chapter_number)
> ```

No need to pass bare field as parameter
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 147 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8sfGj2uuv-TLkd3xP:-OV8sfGj2uuv-TLkd3xQ:b-cw3bpn) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L147)):*
> ```Python
>             ):
>                 self._current_verse_number = scripture_ref.verse_num
>                 self._start_new_verse(self._current_verse_number)
> ```

No need to pass bare field as parameter
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 149 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8sgk2Err7vf1WyO2-:-OV8sgk2Err7vf1WyO20:bei7ss7) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L149)):*
> ```Python
>                 self._start_new_verse(self._current_verse_number)
> 
>     def _start_new_verse(self, new_chapter_number: int) -> None:
> ```

`new_chapter_number` is unused
___
*[`machine/corpora/punctuation_analysis/depth_based_quotation_mark_resolver.py` line 93 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8tFSwAvWrfhdg_NLt:-OV8tFSwAvWrfhdg_NLu:b84supm) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/punctuation_analysis/depth_based_quotation_mark_resolver.py#L93)):*
> ```Python
>         self._quote_continuer_mark_stack.append(quotation_mark)
>         self._continuer_style = quote_continuer_style
>         if len(self._quote_continuer_mark_stack) == len(quotation_mark_resolver_state._quotation_stack):
> ```

Should you use `current_depth + 1`?
___
*[`machine/corpora/punctuation_analysis/depth_based_quotation_mark_resolver.py` line 168 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8tQ4145vqkHfl3nY7:-OV8tQ42DNJ16oSKJcn4:biu4jn9) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/punctuation_analysis/depth_based_quotation_mark_resolver.py#L168)):*
> ```Python
>                 return False
> 
>             # Check the next quotation mark match, since quote continuers must appear consecutively
> ```

Very minor, but some comments are starting with a capital letter - others are lowercase
___
*[`machine/corpora/punctuation_analysis/quote_convention_detector.py` line 53 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8to6YEHbjs_65IcAA:-OV8to6YEHbjs_65IcAB:bpn8m8l) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/punctuation_analysis/quote_convention_detector.py#L53)):*
> ```Python
>         self._quotation_mark_tabulator.tabulate(resolved_quotation_marks)
> 
>     def detect_quote_convention(self, print_summary: bool) -> Union[QuoteConventionAnalysis, None]:
> ```

Throughout the code: Maybe `Optional[Type]` rather than `Union[Type, None]`?
___
*[`machine/corpora/quotation_mark_update_resolution_settings.py` line 16 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8nCft-H_P-JKn0Z-4:-OV8nCft-H_P-JKn0Z-5:bgc0obj) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quotation_mark_update_resolution_settings.py#L16)):*
> ```Python
>         self._source_quote_convention = source_quote_convention
>         self._quote_convention_singleton_set = QuoteConventionSet([self._source_quote_convention])
>         self._target_quote_convention = target_quote_convention
> ```

This is unused.

ddaspit (commented):
> Reviewed 46 of 46 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)* status: all files reviewed, 47 unresolved discussions (waiting on @benjaminking and @Enkidu93)

___
*[`machine/corpora/punctuation_analysis/quote_convention_detector.py` line 53 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8to6YEHbjs_65IcAA:-OVEo6rp7bgcuKCsu9d9:b6znlfk) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/punctuation_analysis/quote_convention_detector.py#L53)):*

Throughout the code: Maybe `Optional[Type]` rather than `Union[Type, None]`?
</blockquote></details>

For Python 3.9 and earlier `Optional[Type]` seems to be the [preferred syntax](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type). Once we move to 3.10+, `Type | None` seems to be preferred.
___
*[`tests/corpora/test_fallback_quotation_mark_resolver.py` line 15 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVEpxUi9mg0uJzw0tYb:-OVEpxUi9mg0uJzw0tYc:b26w7hg) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/tests/corpora/test_fallback_quotation_mark_resolver.py#L15)):*
> ```Python
> 
> def test_reset():
>     english_quote_convention = standard_quote_conventions.STANDARD_QUOTE_CONVENTIONS.get_quote_convention_by_name(
> ```

This constant should probably be exported from the `machine.corpora.punctuation_analysis` package. It would make using it nicer.

benjaminking (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)* status: 21 of 64 files reviewed, 46 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`machine/corpora/quotation_mark_update_first_pass.py` line 37 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8ltuT3mlMCh-C1sji:-OVO-bCb3HdsXSc5RJ0f:b-wcrhcg) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quotation_mark_update_first_pass.py#L37)):*

Some places `depth` is used and elsewhere `level`. Is there a true difference in meaning?
</blockquote></details>

I've tried to use `level` in the context of rules and `depth` in the context of observed quotation marks.  Hopefully this commit clarifies that a bit be
___
*[`machine/corpora/quotation_mark_update_first_pass.py` line 71 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8mTLp8cNYH0LJwJyB:-OVO-swb0guW09LTgGbX:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quotation_mark_update_first_pass.py#L71)):*

Type-hinting for `issues`
</blockquote></details>

Done.
___
*[`machine/corpora/quotation_mark_update_resolution_settings.py` line 16 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8nCft-H_P-JKn0Z-4:-OVO-uI_1S99m0442h0G:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quotation_mark_update_resolution_settings.py#L16)):*

This is unused.
</blockquote></details>

Done.
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 69 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8o0fSDaZulR74AbHn:-OVO-vIk3AJu2K2NWeB1:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L69)):*

Use `Elements`
</blockquote></details>

Done.
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 74 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8o3w13vLEfyNPwv4p:-OVO-vtH1TqMiiF7PrGs:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L74)):*

Use `Elements`
</blockquote></details>

Done.
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 132 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8sDjG-PA6z_8zhgVt:-OVO03V41MxxtJaADL0T:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L132)):*

No need to pass bare field as parameter
</blockquote></details>

Done.
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 147 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8sfGj2uuv-TLkd3xP:-OVO042l5Od_S0bBOzdI:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L147)):*

No need to pass bare field as parameter
</blockquote></details>

Done.
___
*[`machine/corpora/quote_convention_changing_usfm_update_block_handler.py` line 149 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OV8sgk2Err7vf1WyO2-:-OVO04cb6UVM8gg3EqQ3:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/machine/corpora/quote_convention_changing_usfm_update_block_handler.py#L149)):*

`new_chapter_number` is unused
</blockquote></details>

Done.
___
*[`machine/corpora/scripture_embed.py` line  at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBgdMz5_7z_Mz7Xi7o:-OVO0_c11os9J1sHndPX:bvjyiwa) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/scripture_embed.py#L)):*

Is this module used anywhere?
</blockquote></details>

It seems to have been a relic of rebasing.  I've removed it.

Enkidu93 (changes requested):
Sounds good, Ben - thank you. Don't feel obliged to reply in all the Reviewable threads unless there's more to add.

> Reviewed 1 of 49 files at r1, 19 of 46 files at r4, 43 of 43 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OVhei4K3vxk8dVAYzzS:b-1yvtwp)* status: all files reviewed, 6 unresolved discussions (waiting on @benjaminking and @ddaspit)

___
*[`machine/corpora/analysis/quotation_mark_string_match.py` line 57 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaDEYI2-I5_LvWhvGv:-OVhfqd_Bj3VhaE5ffId:b4v3rdc) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_string_match.py#L57)):*

Use `is_at_start_of_segment`?
</blockquote></details>

This is still unaddressed. It's fairly minor though - partially just for consistency with the `next_character` property.
___
*[`machine/corpora/analysis/quote_convention.py` line 122 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUBairZ8rWp7RESO6Ei:-OVhi0J_2MGOwFVDu2Vu:b-4hbkdx) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention.py#L122)):*

I agree.
</blockquote></details>

This one's fixed but not elsewhere
___
*[`machine/corpora/punctuation_analysis/quotation_mark_tabulator.py` line 69 at r5](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVhgKrj17PuaShRZSpf:-OVhgKrj17PuaShRZSpg:b-tyqs6t) ([raw file](https://github.com/sillsdev/machine.py/blob/7e7d577f1ad4a5d7a70e5b4cfbc64fe27cb16574/machine/corpora/punctuation_analysis/quotation_mark_tabulator.py#L69)):*
> ```Python
>         return 1 - (weighted_difference / total_weight)
> 
>     def get_summary_message(self) -> str:
> ```

In general, it looks like the comments Damien and I left regarding overriding `__repr__` or `__str__` have not been addressed. I'm fine with either properly overloading those methods or just removing the code altogether (and supporting methods). @ddaspit, what do you think?

benjaminking (commented):
I rebased again and did a force-push.  Hopefully that fixes the issues with the Git history.

This push also includes a commit to fix a bug that I found with multi-character quotation marks.  This adds a few new tests to cover the corrected behavior.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OVn7TpT9zaNrPlf4eN2:b-itprbl)* status: 45 of 64 files reviewed, 5 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`tests/corpora/test_fallback_quotation_mark_resolver.py` line 15 at r4](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVEpxUi9mg0uJzw0tYb:-OVmmRrq6lahyxlUReZj:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/38d3f0555bfa4b3b09f5402b871b9bfe4ae3b099/tests/corpora/test_fallback_quotation_mark_resolver.py#L15)):*

This constant should probably be exported from the `machine.corpora.punctuation_analysis` package. It would make using it nicer.
</blockquote></details>

Done.
___
*[`machine/corpora/analysis/depth_based_quotation_mark_resolver.py` line 26 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUkqqHW-ptArxzuNwiB:-OVmnX2w3xrwiBGThies:ba8gw3m) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/depth_based_quotation_mark_resolver.py#L26)):*

This property seems like a potential source for errors. Could we settle on one or the other being properly called the current depth and passing the `+/- 1` in the calling code? (...with a comment as needed)
</blockquote></details>

I have changed the depth properties to be consistent with one another.  Depth 1 should always refer to a first-level quotation mark, while a depth of 0 (in the resolver and continuer states) means that no quotation mark has been observed yet.  I think the only time I'm adding one to the depth is when a new quote is opened, which should be a more appropriate use of addition (as opposed to bookkeeping between inconsistent properties).
___
*[`machine/corpora/analysis/quotation_mark_string_match.py` line 57 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaDEYI2-I5_LvWhvGv:-OVn7R1o0hnXO5rrewAU:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quotation_mark_string_match.py#L57)):*

This is still unaddressed. It's fairly minor though - partially just for consistency with the `next_character` property.
</blockquote></details>

Done.
___
*[`machine/corpora/analysis/quote_convention_set.py` line 39 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaF9Ty5MGWna1RAj69:-OVmpTmdBBVvBsdow0UQ:b-n57gnb) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L39)):*

Why are these all sorted (and elsewhere there's similar sorting)? Is it mainly for testing or something like that?
</blockquote></details>

Yes, it's for testing that the regexes are created correctly.  The elements in the set aren't added in a consistent order without using the `sorted` function.
___
*[`machine/corpora/punctuation_analysis/quotation_mark_tabulator.py` line 69 at r5](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVhgKrj17PuaShRZSpf:-OVmpqIyAqRQQXc2lbD2:b-2pyzpd) ([raw file](https://github.com/sillsdev/machine.py/blob/7e7d577f1ad4a5d7a70e5b4cfbc64fe27cb16574/machine/corpora/punctuation_analysis/quotation_mark_tabulator.py#L69)):*

I mainly wanted to get rid of the `print` statements. I will leave it up to Ben to determine if `__repr__`, `__str__`, or a separate method is appropriate. Something like `__str__` or `__repr__` is normally used for short string representations of the object. If the string is a longer more detailed description or represents ancillary information, then a separate method might be more appropriate.
</blockquote></details>

In this particular case, `get_summary_message` creates a message that's useful to communicate to users, but it doesn't fully represent everything that's present in `QuotationMarkTabulator`. It's possible for two meaningfully different instances of `QuotationMarkTabulator` to have identical summary messages.  So it didn't seem like `__str__` or `__repr__` were a good fit for this function.

Enkidu93 (changes requested):
Great - sounds good.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OVnAnDrE0A92_Xq1fmP:b-p43sio)* status: 45 of 64 files reviewed, 2 unresolved discussions (waiting on @benjaminking and @ddaspit)

___
*[`machine/corpora/analysis/depth_based_quotation_mark_resolver.py` line 26 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUkqqHW-ptArxzuNwiB:-OVnApdxCtUUJ8ZZyscT:b-corcx5) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/depth_based_quotation_mark_resolver.py#L26)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

I have changed the depth properties to be consistent with one another.  Depth 1 should always refer to a first-level quotation mark, while a depth of 0 (in the resolver and continuer states) means that no quotation mark has been observed yet.  I think the only time I'm adding one to the depth is when a new quote is opened, which should be a more appropriate use of addition (as opposed to bookkeeping between inconsistent properties).
</blockquote></details>

Yes, sorry - I meant to resolve this comment. This looks good.
___
*[`machine/corpora/analysis/quote_convention_set.py` line 39 at r3](https://reviewable.io/reviews/sillsdev/machine.py/203#-OUaF9Ty5MGWna1RAj69:-OVnCgKO1CVDl-FuxcRp:bs0e4l7) ([raw file](https://github.com/sillsdev/machine.py/blob/83d89b80fdcc4c987eef4cb9196e5b9cab2ab821/machine/corpora/analysis/quote_convention_set.py#L39)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

Yes, it's for testing that the regexes are created correctly.  The elements in the set aren't added in a consistent order without using the `sorted` function.
</blockquote></details>

Yes - sorry - meant to resolve this. Makes sense. Something to beware of is that C# and Python do not always sort strings in the same way (oddly enough), so it will create some discrepancies in the tests (which other approaches like making both strings into sets of chars in the test itself wouldn't). But it's not a big deal. 
___
*[`machine/corpora/punctuation_analysis/quotation_mark_tabulator.py` line 69 at r5](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVhgKrj17PuaShRZSpf:-OVnB-i05Az_8PvtChHQ:b14todt) ([raw file](https://github.com/sillsdev/machine.py/blob/7e7d577f1ad4a5d7a70e5b4cfbc64fe27cb16574/machine/corpora/punctuation_analysis/quotation_mark_tabulator.py#L69)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

In this particular case, `get_summary_message` creates a message that's useful to communicate to users, but it doesn't fully represent everything that's present in `QuotationMarkTabulator`. It's possible for two meaningfully different instances of `QuotationMarkTabulator` to have identical summary messages.  So it didn't seem like `__str__` or `__repr__` were a good fit for this function.
</blockquote></details>

OK, I would say that if the main concern is that it doesn't capture all instance-specific info needed to recreate the object, then that would still be an appropriate place to use `__str__` just not `__repr__`. But if you think it's too long or can't be reworked to make sense as a user-friendly string representation but is more like secondary, diagnostic info, then that's fine. I mainly am just trying to avoid committing code unnecessarily - especially since this function requires other methods that are only used here. Is there a reason that you believe this class in particular benefits from a summarizing function like this? This class isn't really user-facing, correct?

benjaminking (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203)* status: 43 of 64 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`machine/corpora/punctuation_analysis/quotation_mark_tabulator.py` line 69 at r5](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVhgKrj17PuaShRZSpf:-OVns52j68yEbhUmbAj4:b-3d99gk) ([raw file](https://github.com/sillsdev/machine.py/blob/7e7d577f1ad4a5d7a70e5b4cfbc64fe27cb16574/machine/corpora/punctuation_analysis/quotation_mark_tabulator.py#L69)):*

OK, I would say that if the main concern is that it doesn't capture all instance-specific info needed to recreate the object, then that would still be an appropriate place to use `__str__` just not `__repr__`. But if you think it's too long or can't be reworked to make sense as a user-friendly string representation but is more like secondary, diagnostic info, then that's fine. I mainly am just trying to avoid committing code unnecessarily - especially since this function requires other methods that are only used here. Is there a reason that you believe this class in particular benefits from a summarizing function like this? This class isn't really user-facing, correct?
</blockquote></details>

Your comment made me realize that I still had a print statement that was using this function.  The message is one that the EITL team found helpful to include in the standalone script that detected the quotation convention.  I made a change to include it in the `QuoteConventionAnalysis` object, so that the script can choose whether or not to print it.  With that new structure in place, I think it is now clearer to leave the function as `get_summary_message` than to change it into `__str__`.

benjaminking (commented):
Two more small changes:
1) a check for an edge case with quote continuers that could lead to an out-of-bounds error
2) A little bit of logic to create better guesses for ambiguous quotation marks in rare situations

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OW2VetHAdkmomvXdtUH:bgkq6f)* status: 40 of 64 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @Enkidu93)

Enkidu93 (changes requested):
Great! Thank you

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OWHmQgD5uAfhw_5J52p:bv10ema)* status: 40 of 64 files reviewed, 9 unresolved discussions (waiting on @benjaminking and @ddaspit)

___
*[`machine/corpora/punctuation_analysis/quotation_mark_tabulator.py` line 69 at r5](https://reviewable.io/reviews/sillsdev/machine.py/203#-OVhgKrj17PuaShRZSpf:-OWHmU8--haYBgTq1OqV:b-hfghm4) ([raw file](https://github.com/sillsdev/machine.py/blob/7e7d577f1ad4a5d7a70e5b4cfbc64fe27cb16574/machine/corpora/punctuation_analysis/quotation_mark_tabulator.py#L69)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

Your comment made me realize that I still had a print statement that was using this function.  The message is one that the EITL team found helpful to include in the standalone script that detected the quotation convention.  I made a change to include it in the `QuoteConventionAnalysis` object, so that the script can choose whether or not to print it.  With that new structure in place, I think it is now clearer to leave the function as `get_summary_message` than to change it into `__str__`.
</blockquote></details>

Sounds good
___
*[`tests/corpora/punctuation_analysis/test_quotation_mark_string_match.py` line 157 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHiDxz5Iil6f-d8gfe:-OWHiDxz5Iil6f-d8gff:bik79kf) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/tests/corpora/punctuation_analysis/test_quotation_mark_string_match.py#L157)):*
> ```Python
> 
>     quotation_mark_string_match = QuotationMarkStringMatch(
>         TextSegment.Builder().set_text("sample text").build(), 11, 12
> ```

I'm confused what an end index of twelve would mean here? It doesn't change the substance of the test though. 
___
*[`tests/corpora/punctuation_analysis/test_depth_based_quotation_mark_resolver.py` line 996 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHhOQP6vJYd3ObTApi:-OWHhOQQ-cQ7HHkbkx7U:b-lgffbj) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/tests/corpora/punctuation_analysis/test_depth_based_quotation_mark_resolver.py#L996)):*
> ```Python
>     )
>     assert not standard_swedish_quotation_mark_categorizer.is_opening_quotation_mark(
>         QuotationMarkStringMatch(TextSegment.Builder().set_text("\u201d").build(), 1, 2)
> ```

I think you meant `...0, 1)` not `...1, 2)`
___
*[`tests/corpora/punctuation_analysis/test_depth_based_quotation_mark_resolver.py` line 2293 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHglwJ2XMXxBEp8Xjy:-OWHglwJ2XMXxBEp8Xjz:ba6xw9b) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/tests/corpora/punctuation_analysis/test_depth_based_quotation_mark_resolver.py#L2293)):*
> ```Python
>         QuotationMarkMetadata("\u201c", 3, QuotationMarkDirection.OPENING, text_segment, 10, 11),
>         QuotationMarkMetadata("\u2018", 4, QuotationMarkDirection.OPENING, text_segment, 13, 14),
>         # QuotationMarkMetadata("\u201c", 5, QuotationMarkDirection.Opening, text_segment, 20, 21),
> ```

Can this be removed?
___
*[`tests/corpora/punctuation_analysis/test_quote_convention.py` line 370 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHhiA1B3rofVZZxVZv:-OWHhiA1B3rofVZZxVZw:bkbhz44) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/tests/corpora/punctuation_analysis/test_quote_convention.py#L370)):*
> ```Python
> def test_print_summary() -> None:
>     quote_convention = QuoteConvention(
>         "test-quote-convention",
> ```

In this class - not sure if it's elsewhere - you're using kebab-case for quote convention names rather than the snake_case which is used elsewhere.
___
*[`tests/corpora/test_quotation_denormalization.py` line 53 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHgQwR7CmaI8RToSSB:-OWHgQwR7CmaI8RToSSC:b-i23nby) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/tests/corpora/test_quotation_denormalization.py#L53)):*
> ```Python
>     actual_denormalized_usfm = updater.get_usfm()
> 
>     assert actual_denormalized_usfm == expected_denormalized_usfm
> ```

For consistency, you can use the `ignore_line_endings` function in the helper class to avoid having to stick an `\r` at each line-ending
___
*[`tests/corpora/punctuation_analysis/test_text_segment.py` line 75 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHlmGZ3JLSMkGYqkXc:-OWHlmGZ3JLSMkGYqkXd:bb7j6qk) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/tests/corpora/punctuation_analysis/test_text_segment.py#L75)):*
> ```Python
>     different_text_segment = TextSegment.Builder().set_text("different text").build()
> 
>     assert basic_segment == basic_segment
> ```

This line and then following line don't seem to be what you meant to put
___
*[`machine/corpora/quotation_mark_update_first_pass.py` line 43 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHflhr2bfHaTLBoTsc:-OWHflhr2bfHaTLBoTsd:bozj7dw) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/machine/corpora/quotation_mark_update_first_pass.py#L43)):*
> ```Python
>             if depth <= target_quote_convention.num_levels:
>                 target_marks_by_source_marks[opening_quotation_mark].add(
>                     target_quote_convention.get_closing_quotation_mark_at_depth(depth)
> ```

This loop could be cut short - i.e., you shouldn't need to re-iterate.

Enkidu93:
*[`machine/corpora/quotation_mark_update_first_pass.py` line 43 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHflhr2bfHaTLBoTsc:-OWW-_0I6s07GnZJe7E-:b84duq3) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/machine/corpora/quotation_mark_update_first_pass.py#L43)):*

That is an improvement but not what I was thinking. Is there a reason something like this wouldn't work?
```
        source_marks: set = set()
        for depth in range(1, min(source_quote_convention.num_levels, target_quote_convention.num_levels) + 1):
            opening_quotation_mark = source_quote_convention.get_opening_quotation_mark_at_depth(depth)
            if opening_quotation_mark in source_marks:
                return False
            source_marks.add(opening_quotation_mark)
        return True
```
</blockquote></details>

Sorry, more like this, I guess:
```
        target_mark_by_source_mark: Dict[str, str] = {}
        for depth in range(1, min(source_quote_convention.num_levels, target_quote_convention.num_levels) + 1):
            opening_quotation_mark = source_quote_convention.get_opening_quotation_mark_at_depth(depth)
            closing_quotation_mark = target_quote_convention.get_closing_quotation_mark_at_depth(depth)
            if (
                opening_quotation_mark in target_mark_by_source_mark
                and target_mark_by_source_mark[opening_quotation_mark] != closing_quotation_mark
            ):
                return False
            target_mark_by_source_mark[opening_quotation_mark] = closing_quotation_mark
        return True
```

Enkidu93 (approved):
:lgtm: 

Well done! It's been an adventurous code review. Thank you for all your hard work :heart: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/203#-:-OWXQMGW243yZ0YAVGmg:b-be05)* status: 34 of 64 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`machine/corpora/quotation_mark_update_first_pass.py` line 43 at r7](https://reviewable.io/reviews/sillsdev/machine.py/203#-OWHflhr2bfHaTLBoTsc:-OWXQEfK4j14cFIUfFG-:btojbfd) ([raw file](https://github.com/sillsdev/machine.py/blob/f93a19566877530ef6a335ec0112086ef01630fd/machine/corpora/quotation_mark_update_first_pass.py#L43)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

Yes, that should work.  Incidentally, your comment made me realize that there was actually a bug in this function.  It should be comparing opening quotes to opening quotes and closing quotes to closing quotes, instead of comparing opening to closing quotes.  The tests didn't catch it earlier since real quote conventions tend to be parallel across levels.  I fixed the bug (using the function structure you suggested) and added a test that will fail for the old logic, but pass for the new logic.
</blockquote></details>

Great!

ddaspit:
Don't forget to cleanup/squash the commits when merging.

PR #224, opened 2025-09-04
Detect quote convention from file Paratext project

benjaminking:
This PR adds a class that can detect the quote convention for a Paratext project from a project directory.  This class will primarily be used by SILNLP to detect the quote convention for the source and target projects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/224)

### DIFF
--- a/machine/corpora/file_paratext_project_quote_convention_detector.py
+++ b/machine/corpora/file_paratext_project_quote_convention_detector.py
@@ -1,0 +1,19 @@
+from pathlib import Path
+from typing import BinaryIO, Optional
+
+from ..utils.typeshed import StrPath
+from .file_paratext_project_settings_parser import FileParatextProjectSettingsParser
+from .paratext_project_quote_convention_detector import ParatextProjectQuoteConventionDetector
+
+
+class FileParatextProjectQuoteConventionDetector(ParatextProjectQuoteConventionDetector):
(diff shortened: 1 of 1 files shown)


Enkidu93 (changes requested):
> @Enkidu93 reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/224)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`machine/corpora/file_paratext_project_quote_convention_detector.py` line 18 at r1](https://reviewable.io/reviews/sillsdev/machine.py/224#-OZP2zrDBgUoc0OhDxzT:-OZP2zrDBgUoc0OhDxzU:b-dt480j) ([raw file](https://github.com/sillsdev/machine.py/blob/ccdbbfbbc0955d7ba9b843347fdec2bb2b0f1863/machine/corpora/file_paratext_project_quote_convention_detector.py#L18)):*
> ```Python
>         return (Path(self._project_dir) / file_name).exists()
> 
>     def _open(self, file_name: str) -> Optional[BinaryIO]:
> ```

I don't think this method can ever return `None`, so you don't need `Optional`. Also, you should use `StrPath` rather than `str` consistently here to match the other `file_...` implementation classes.

PR #226, opened 2025-09-09
Detect quote convention for file pt proj (minor fix)

benjaminking:
Unfortunately I forgot to include `FileParatextProjectQuoteConventionDetector` in `__init__.py`, so the released version is not usable in SILNLP for quote convention detection.

(By the way, is it better to create a new branch in cases like this or just modify the already merged branch?)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/226)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ reportMissingModuleSource = false
 
 [tool.poetry]
 name = "sil-machine"
-version = "1.8.1"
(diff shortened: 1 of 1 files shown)


Enkidu93:
I went ahead and rebased it. Like I said, if you just use a new branch, then you won't run into merge conflict issues like these (even if they're easy to resolve).

Issue #222, opened 2025-08-28
Port test cases added in Machine to cover C# Unicode-handling


Enkidu93:
Tests added in https://github.com/sillsdev/machine/pull/327 and in the PR related to the issue mentioned there.

PR #228, opened 2025-09-11
Port quotation denormalization unicode tests

Enkidu93:
Addresses https://github.com/sillsdev/machine.py/issues/222

Added the same tests.

Unfortunately, using the `TextElement` strategy in Machine means that combining characters are joined into preceding text elements - something which Python does not do. I updated the code to handle them the same as Machine. The alternative is just to allow there to be a difference. It's mostly internal, so it may not be an issue to do so.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/228)


Enkidu93:
> I think I would prefer to target the specific issue, i.e. surrogate pairs in C#, and do no more. So let's not use `StringInfo.GetTextElementEnumerator` and make our own implementation that properly deals with surrogate pairs.
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/228#-:-O_DcPB0-aLWsAUVP0eF:bygiggi)_ status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Sounds good. I've removed the custom code here and will add commits to the parallel Machine PR to address this issue there.

PR #170, opened 2025-04-23
Fix handling of implicitly closed char styles when updating USFM

ddaspit:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/170)

PR #237, opened 2025-10-02
Port machine changes

Enkidu93:
Port 
* https://github.com/sillsdev/machine/pull/343
* https://github.com/sillsdev/machine/pull/342
* https://github.com/sillsdev/machine/pull/341

Sorry this took so long - I hit a bug that I just couldn't figure out since I was certain I had ported correctly. Turns out that machine.py and Machine code had become slightly out of sync. I went ahead and made the change here to make it consistent with Machine, but I'm suspicious that Machine was actually the incorrect one/missing a commit. I think that maybe https://github.com/sillsdev/machine.py/pull/170 never got entirely properly incorporated into Machine because it went in around the same time as the update block code. Could you take a peek, @ddaspit, and let me know what's right and what's wrong? Also happy to just push this through since the tests are passing and follow up on this in a separate issue.

### DIFF
--- a/machine/corpora/paratext_project_terms_parser_base.py
+++ b/machine/corpora/paratext_project_terms_parser_base.py
@@ -70,10 +70,14 @@ class ParatextProjectTermsParserBase(ABC):
                 id = term.attrib["Id"]
                 if _is_in_category(id, term_categories, term_id_to_category_dict):
                     id_ = id.replace("\n", "&#xA")
-                    renderings = term.find("Renderings")
-                    gloss = renderings.text if renderings is not None and renderings.text is not None else ""
-                    glosses = _get_glosses(gloss)
-                    terms_renderings[id_].extend(glosses)
+                    renderings_element = term.find("Renderings")
+                    rendering_text = (
+                        renderings_element.text
+                        if renderings_element is not None and renderings_element.text is not None
+                        else ""
+                    )
+                    renderings = _get_renderings(rendering_text)
+                    terms_renderings[id_].extend(renderings)
 
         terms_glosses: Dict[str, List[str]] = defaultdict(list)
         if terms_glosses_doc is not None and use_term_glosses:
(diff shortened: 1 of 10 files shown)


ddaspit (changes requested):
> @ddaspit reviewed 8 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/237)* status: 8 of 10 files reviewed, 1 unresolved discussion

___
*[`machine/corpora/usfm_update_block_handler.py` line 11 at r1](https://reviewable.io/reviews/sillsdev/machine.py/237#-OaeM4XG9iG2Rp3P_o1E:-OaeM4XG9iG2Rp3P_o1F:b-2qdsul) ([raw file](https://github.com/sillsdev/machine.py/blob/9618d8f8ac2ea1532c2604fc3023f5dbfd5b2490/machine/corpora/usfm_update_block_handler.py#L11)):*
> ```Python
> 
> 
> class UsfmUpdateBlockHandlerException(Exception):
> ```

You should name this `UsfmUpdateBlockHandlerError` so that it is consistent with Python naming conventions.

ddaspit (commented):
> @ddaspit reviewed 2 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/237)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`machine/corpora/scripture_ref.py` line 125 at r1](https://reviewable.io/reviews/sillsdev/machine.py/237#-OafLfKaEXlPNBuy6Kv3:-OafLfKaEXlPNBuy6Kv4:b3nnb6m) ([raw file](https://github.com/sillsdev/machine.py/blob/9618d8f8ac2ea1532c2604fc3023f5dbfd5b2490/machine/corpora/scripture_ref.py#L125)):*
> ```Python
>         # Using to_relaxed() is necessary to maintain equality across relaxed refs,
>         # __eq__ properly handles relaxed ref comparison
>         return hash((self.verse_ref, tuple(self.to_relaxed().path)))
> ```

This is different than the implementation in Machine. On the other hand, Machine is inconsistent with itself, which is probably not a good thing. `ScriptureRefComparer.GetHashCode()` and `ScriptureRef.GetHashCode()` give different results, which would be unexpected for a caller. Why do we need to call `to_relaxed` here?

ddaspit (commented):
> @ddaspit reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/237)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

PR #239, opened 2025-10-30
Improved quote convention detection for Paratext projects

benjaminking:
This PR improves quote convention detection for Paratext projects, especially messy projects that are inconsistent with their quote conventions.  It does this by implementing a weighted voting scheme across different books.  It also makes a small change to the way that quote convention similarity is calculated to accommodate the weighted voting.  Finally, it adds a new quote convention that was recently observed in a project.

On a set of 57 real projects submitted to Serval, this improves the accuracy of quote convention detection from 40% to 95%.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/239)

### DIFF
--- a/machine/punctuation_analysis/paratext_project_quote_convention_detector.py
+++ b/machine/punctuation_analysis/paratext_project_quote_convention_detector.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from collections import defaultdict
 from typing import BinaryIO, Dict, List, Optional, Union
 
 from ..corpora.paratext_project_settings import ParatextProjectSettings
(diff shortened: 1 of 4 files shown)


ddaspit (changes requested):
> @ddaspit reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/239)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`machine/punctuation_analysis/quote_convention_detector.py` line 53 at r1](https://reviewable.io/reviews/sillsdev/machine.py/239#-OdAcX1j32NXGDbpofJ-:-OdAcX1j32NXGDbpofJ0:bx3rn6a) ([raw file](https://github.com/sillsdev/machine.py/blob/afa93ba970a7a7fe7b533dca0caffd156016a3c3/machine/punctuation_analysis/quote_convention_detector.py#L53)):*
> ```Python
>         return STANDARD_QUOTE_CONVENTIONS.score_all_quote_conventions(self._quotation_mark_tabulator)
> 
>     def detect_quote_convention_and_get_tabulated_quotation_marks(
> ```

Could we expose the tabulated quotation marks from the `QuoteConventionAnalysis` instead of returning it separately?
___
*[`machine/punctuation_analysis/quote_convention.py` line 64 at r1](https://reviewable.io/reviews/sillsdev/machine.py/239#-OdAbhsZAMbuSDB_19C0:-OdAbhsZAMbuSDB_19C1:b-ykmwre) ([raw file](https://github.com/sillsdev/machine.py/blob/afa93ba970a7a7fe7b533dca0caffd156016a3c3/machine/punctuation_analysis/quote_convention.py#L64)):*
> ```Python
> 
>     def __hash__(self) -> int:
>         return hash((tuple(self.level_conventions)))
> ```

Are the extra parentheses necessary? Doesn't `tuple` return a tuple?

benjaminking:
For some reason, Reviewable is crashing every time I open this PR, so I'm just going to respond on Github.



> Could we expose the tabulated quotation marks from the `QuoteConventionAnalysis` instead of returning it separately?

I have refactored this to avoid having to either expose additional fields or return multiple values from a method.



> Are the extra parentheses necessary? Doesn't `tuple` return a tuple?

Yes, I have removed the extraneous parentheses.



> I'm noticing that these scores all seem to be going down. Why is that the case? If the intent is to use the same logic as before but weight it by book, shouldn't these stay the same? Is it because of this: `# The scores of greater depths depend on the scores of shallower depths`? Is the idea that if the top-level quotes are off, that should be compounded into the score for deeper quotes? Was this motivated by particular examples?

Yes, the changes in score are due to the way that the quotation matching score is calculated (as you mentioned).  Yes, the idea is that we shouldn't give credit for matching 2nd-level quotation marks unless the 1st-level quotation marks match (to some extent).  There was a situation that was happening in a few cases that motivated this.

Suppose that a project has some books that use convention X and some that use convention Y.  Now suppose that convention Z has the same 1st-level quotation marks as X and the same 2nd-level quotation marks as Y.  In a lot of cases, Z was being identified as the quote convention for the project, even though none of the books used Z.  So I made the change to only give a convention credit for having matching 2nd-level quotation marks to the extent that it also matches the 1st-level quotation marks.

There should also be a number of situations where the score will increase under the new calculation.



> We aren't thresholding on these values at the moment anywhere, are we? If so, we need to make sure we update those threshold values.

I don't believe we have set any thresholds based on this score.

> You should combine this with setting `self._best_quote_convention_score` so you don't have to calculate `max(...)` twice.

Done.

Enkidu93 (approved):
Reviewable is also crashing for me :/.

Great, looks good. Thank you for the details on the depth-based adjustment.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/239#-:-OdjSpMaAzQalhXlYlV8:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

Issue #159, opened 2025-02-26
Implement `NParallelTextCorpus`


Enkidu93:
Since `NParallelTextCorpus` in Machine (C#) now houses the implementation for `ParallelTextCorpus.GetRows()` which undergirds text alignment, this is particular important.

johnml1135:
Is this to align SILNLP's mixed source with Serval's?  It makes sense - I just want to understand the relative priority.

PR #247, opened 2025-12-01
Port `NParallelText` corpus

Enkidu93:
Fixes https://github.com/sillsdev/machine.py/issues/159

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/247)

### DIFF
--- a/machine/corpora/standard_parallel_text_corpus.py
+++ b/machine/corpora/standard_parallel_text_corpus.py
@@ -27,8 +27,7 @@ class StandardParallelTextCorpus(ParallelTextCorpus):
         self._all_source_rows = all_source_rows
         self._all_target_rows = all_target_rows
         self._n_parallel_text_corpus = NParallelTextCorpus([source_corpus, target_corpus])
-        self._n_parallel_text_corpus.all_rows[0] = self.all_source_rows
-        self._n_parallel_text_corpus.all_rows[1] = self.all_target_rows
+        self._n_parallel_text_corpus.all_rows[:] = [self.all_source_rows, self.all_target_rows]
         self._row_ref_comparer = row_ref_comparer or default_row_ref_comparer
 
     @property
--- a/tests/corpora/test_n_parallel_text_corpus.py
+++ b/tests/corpora/test_n_parallel_text_corpus.py
@@ -115,7 +115,7 @@ def test_get_rows_three_corpora_missing_rows_all_all_rows():
         )
     )
     n_parallel_corpus = NParallelTextCorpus([corpus1, corpus2, corpus3])
-    n_parallel_corpus._all_rows = [True, True, True]
+    n_parallel_corpus.all_rows[:] = [True, True, True]
     rows = list(n_parallel_corpus.get_rows())
     assert len(rows) == 3
     assert all([r[0] == 3 for r in rows[2].n_refs])
@@ -152,7 +152,7 @@ def test_get_rows_three_corpora_missing_rows_some_all_rows():
         )
     )
     n_parallel_corpus = NParallelTextCorpus([corpus1, corpus2, corpus3])
-    n_parallel_corpus._all_rows = [True, False, True]
+    n_parallel_corpus.all_rows[:] = [True, False, True]
     rows = list(n_parallel_corpus.get_rows())
     assert len(rows) == 2
     assert all([r[0] == 3 for r in rows[1].n_refs])
@@ -192,7 +192,7 @@ def test_get_rows_three_corpora_missing_rows_all_all_rows_missing_middle():
         )
     )
     n_parallel_corpus = NParallelTextCorpus([corpus1, corpus2, corpus3])
-    n_parallel_corpus._all_rows = [True, True, True]
+    n_parallel_corpus.all_rows[:] = [True, True, True]
     rows = list(n_parallel_corpus.get_rows())
     assert len(rows) == 3
     assert all([len(r) == 0 or r[0] == 2 for r in rows[1].n_refs])
@@ -228,7 +228,7 @@ def test_get_rows_three_corpora_missing_rows_missing_last_rows():
         )
     )
     n_parallel_corpus = NParallelTextCorpus([corpus1, corpus2, corpus3])
-    n_parallel_corpus._all_rows = [True, False, False]
+    n_parallel_corpus.all_rows[:] = [True, False, False]
     rows = list(n_parallel_corpus.get_rows())
     assert len(rows) == 3
     assert all([r[0] == 2 for r in rows[1].n_refs])
@@ -247,7 +247,7 @@ def test_get_rows_three_corpora_one_corpus():
         )
     )
     n_parallel_corpus = NParallelTextCorpus([corpus1])
-    n_parallel_corpus._all_rows = [True]
+    n_parallel_corpus.all_rows[:] = [True]
     rows = list(n_parallel_corpus.get_rows())
     assert len(rows) == 2
     assert all([r[0] == 1 for r in rows[0].n_refs])
@@ -389,7 +389,7 @@ def test_get_rows_three_corpora_overlapping_ranges_all_individual_rows():
         )
     )
     n_parallel_corpus = NParallelTextCorpus([corpus1, corpus2, corpus3])
-    n_parallel_corpus._all_rows = [False, False, True]
+    n_parallel_corpus.all_rows[:] = [False, False, True]
     rows = list(n_parallel_corpus.get_rows())
     assert len(rows) == 3
     assert rows[0].n_refs[0] == [1]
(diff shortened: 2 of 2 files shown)


ddaspit (changes requested):
> @ddaspit reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/247)* status: all files reviewed, 6 unresolved discussions (waiting on @Enkidu93)

___
*[`machine/corpora/n_parallel_text_corpus.py` line 84 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVYYPb-YaeCRqyD2mo:-OfVYYPc1wr_JL1YH1Dn:b-y3mn8) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus.py#L84)):*
> ```Python
>         self._corpora = corpora
>         self._row_ref_comparer = row_ref_comparer if row_ref_comparer is not None else default_row_ref_comparer
>         self.all_rows = [False for _ in range(len(corpora))]
> ```

The type should be `Sequence[bool]`.

Also, I know that you are porting this as-is from the C#, but I don't like that this property is mutable.
___
*[`machine/corpora/n_parallel_text_corpus.py` line 94 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVXxTTBkmS5O-jpC0H:-OfVXxTTBkmS5O-jpC0I:bsw8u8q) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus.py#L94)):*
> ```Python
> 
>     @property
>     def corpora(self) -> List[TextCorpus]:
> ```

The type should be `Sequence[TextCorpus]`.
___
*[`machine/corpora/n_parallel_text_corpus.py` line 380 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfV_NgK6N2Gdv83kv_H:-OfV_NgK6N2Gdv83kv_I:b-g79mzc) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus.py#L380)):*
> ```Python
> 
> 
> def default_row_ref_comparer(x: object, y: object) -> int:
> ```

I think the type for `x` and `y` should be `Any`.
___
*[`machine/corpora/text_corpus.py` line 31 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVTuKW0fXmM_M0ZytS:-OfVTuKW0fXmM_M0ZytT:b-c1ap7u) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/text_corpus.py#L31)):*
> ```Python
>     @property
>     @abstractmethod
>     def versification(self) -> Versification: ...
> ```

Please change the type here to `Optional[Versification]`, since it is now clear that it can be `None`.
___
*[`machine/corpora/n_parallel_text_corpus_base.py` line 20 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVWeiA0cidmZBkv_YG:-OfVWeiB12RY-ycpeKFW:b9i31u8) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus_base.py#L20)):*
> ```Python
> 
>     @abstractmethod
>     def get_rows(self, text_ids: List[str]) -> Sequence[NParallelTextRow]: ...
> ```

The type for `text_ids` should be `Iterable[str]`.
___
*[`machine/corpora/text_corpus_enumerator.py` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVUPbXCJ61LunQMUWm:-OfVUPbXCJ61LunQMUWn:b19bzr5) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/text_corpus_enumerator.py#L9)):*
> ```Python
> 
> 
> class _TextCorpusEnumerator(ContextManager["_TextCorpusEnumerator"], Generator[TextRow, None, None]):
> ```

This should be named `TextCorpusEnumerator`, since it is used outside of this module.

ddaspit (commented):
> @ddaspit reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/247)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`machine/corpora/n_parallel_text_corpus.py` line 84 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVYYPb-YaeCRqyD2mo:-Of_aGOf34yq6eTmwGzl:b-svw7id) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus.py#L84)):*

Done.

I've made it a tuple, but I'm not sure what solution you have in mind. These are just default values here anyways. I could allow `all_rows` to be passed through the constructor and then make this property 'private' entirely.
</blockquote></details>

What I don't like is that anything can just set `all_rows` to a new value, even a value with the wrong length. `all_rows` should be read-only, but boolean values inside can still be mutable. Does that make sense?
___
*[`machine/corpora/n_parallel_text_row.py` line 21 at r2](https://reviewable.io/reviews/sillsdev/machine.py/247#-Of_pVEv76oY3znQJZLn:-Of_pVEv76oY3znQJZLo:b-cth1qi) ([raw file](https://github.com/sillsdev/machine.py/blob/def8b0a59832d68d8c03fd57c7b91d9e2c87686e/machine/corpora/n_parallel_text_row.py#L21)):*
> ```Python
> 
>     @property
>     def ref(self) -> object:
> ```

You should use `Any` here. That is the type that `ParallelTextRow` and `TextRow` use for refs. I would just use `Any` as the type for refs everywhere to be consistent across the codebase.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/247)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[`machine/corpora/n_parallel_text_corpus.py` line 84 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVYYPb-YaeCRqyD2mo:-Of_vZDwCuG_mnRaE9Rg:b1pkun7) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus.py#L84)):*

What I don't like is that anything can just set `all_rows` to a new value, even a value with the wrong length. `all_rows` should be read-only, but boolean values inside can still be mutable. Does that make sense?
</blockquote></details>

OK, so you do want it to be mutable, just read-only. Makes sense. Done. Does this have the effect you're looking for? Is it OK to leave as-is in the tests or should I set them index-by-index there as well?
___
*[`machine/corpora/n_parallel_text_row.py` line 21 at r2](https://reviewable.io/reviews/sillsdev/machine.py/247#-Of_pVEv76oY3znQJZLn:-Of_vgnLA8JT1n4dMfgU:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/def8b0a59832d68d8c03fd57c7b91d9e2c87686e/machine/corpora/n_parallel_text_row.py#L21)):*

You should use `Any` here. That is the type that `ParallelTextRow` and `TextRow` use for refs. I would just use `Any` as the type for refs everywhere to be consistent across the codebase.
</blockquote></details>

Done.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/247)* status: 8 of 10 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`machine/corpora/n_parallel_text_corpus.py` line 84 at r1](https://reviewable.io/reviews/sillsdev/machine.py/247#-OfVYYPb-YaeCRqyD2mo:-OfeI4oT1qX4m8Kwr_ZO:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/b4e079f06df230d0885f4e431ce3300602bbf725/machine/corpora/n_parallel_text_corpus.py#L84)):*

I would prefer that you don't set the underlying `_all_rows` field. I think you can set all of the values in one call, like this: `all_rows[:] = [true, true]`.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/247#-:-OfipTI9DZ50fznMteVC:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #257, opened 2026-01-02
Key terms updates necessary for use in SILNLP

Enkidu93:
Added support for capturing renderings patterns, references, and term domains. Moved to using a `KeyTerm` data structure rather than tuples. 

(This also includes porting of recent changes in Machine https://github.com/sillsdev/machine/pull/362 and https://github.com/sillsdev/machine/pull/368)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/257)

### DIFF
--- a/machine/corpora/scripture_text.py
+++ b/machine/corpora/scripture_text.py
@@ -12,7 +12,7 @@ from .text_row_content_type import TextRowContentType
 
 class ScriptureText(TextBase):
     def __init__(self, id: str, versification: Optional[Versification] = None) -> None:
-        super().__init__(id, get_scripture_text_sort_key(id), content_type=TextRowContentType.SEGMENT)
+        super().__init__(id, get_scripture_text_sort_key(id), default_content_type=TextRowContentType.SEGMENT)
         self._versification = ENGLISH_VERSIFICATION if versification is None else versification
 
     @property
--- a/machine/corpora/text_base.py
+++ b/machine/corpora/text_base.py
@@ -6,10 +6,12 @@ from .text_row_content_type import TextRowContentType
 
 
 class TextBase(Text):
-    def __init__(self, id: str, sort_key: str, content_type: TextRowContentType = TextRowContentType.SEGMENT) -> None:
+    def __init__(
+        self, id: str, sort_key: str, default_content_type: TextRowContentType = TextRowContentType.SEGMENT
+    ) -> None:
         self._id = id
         self._sort_key = sort_key
-        self._default_content_type = content_type
+        self._default_content_type = default_content_type
 
     @property
     def id(self) -> str:


Enkidu93:
Also, update machine.py library version

ddaspit (changes requested):
> @ddaspit partially reviewed 11 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/257)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93).

___
*[`machine/corpora/key_term_row.py` line 0 at r1](https://reviewable.io/reviews/sillsdev/machine.py/257#-OiJGNaOBCr4BAYv6i4O:-OiJGNaOBCr4BAYv6i4P:b-ep5fnt) ([raw file](https://github.com/sillsdev/machine.py/blob/750599f621a0a86eae4c079d9978230e5122706b/machine/corpora/key_term_row.py#L0)):*
This file should be named `key_term.py`.

Enkidu93 (commented):
> @Enkidu93 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/257)* status: 17 of 26 files reviewed, 1 unresolved discussion (waiting on @ddaspit).

___
*[`machine/translation/huggingface/hugging_face_nmt_model_trainer.py` line 94 at r4](https://reviewable.io/reviews/sillsdev/machine.py/257#-Oj1_iRD5v0rnNsKo7aN:-OjLs5Uw-s0ZccSplCJc:b-1u3klv) ([raw file](https://github.com/sillsdev/machine.py/blob/c4d0b8b60c009627175a8f1a589b3a11de004aa1/machine/translation/huggingface/hugging_face_nmt_model_trainer.py#L94)):*

I think it would be better to add some type of metadata to a row to indicate what kind of data it is, rather than pass around two separate corpora. It feels more in spirit with the design goals of the corpus API in Machine. We could have a generic metadata dictionary that allows you to set whatever you want. Or, we could add a new "data type" property with a predefined set of values, such as gloss, sentence, section, document, etc. What do you think?
</blockquote></details>

Yes, I did think of this. If we wanted to add a field to each row, I think an enum like you mentioned might be the better option. We could add it to each text and then expose it through the row as well (?). There were two reasons I didn't already tackle this:

1. It will add a little complexity to the `to_hf_dataset()` function as well as to the `preprocess_function` when we call `map()` on the dataset.
2. I wondered how much it was worth doing this when we only have two categories. Do we foresee additional text types? Or maybe we don't foresee other text types but do we see ourselves needing some kind of tagging for other attributes? This would affect how it should be implemented.

A competing (or perhaps supplemental) option would be for this function to take something like `corpora: List[Tuple[CorpusOptions, Union[ParallelCorpus, Dataset]]]` or `corpora: List[CorpusConfig]`. Then the processing options could be more dynamic - e.g., 'include partial words for corpus 1, tag corpus 2 with such-and-such tag, etc.' - regardless of a text type field and it could be specified by the calling code.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/257#-:-OjbFai7AHQ6VVIanaZG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93).

PR #285, opened 2026-03-26
modify usfm for chapter-level drafting to avoid import issues; move remarks to chapters

mshannon-sil:
This PR addresses issue #284.

Mostly looking for high-level feedback about the approach at the moment. As we were discussing, is the right place for this functionality in the `get_usfm()` method as essentially a post-processing step? Or should we look to implement this feature in `process_tokens()` (and maybe move the remark logic here as well)?

Some initial thoughts:
Pros for putting it in `get_usfm()`:
- The code is together in a cohesive unit making it potentially easier to maintain, rather than spread across `process_token()`.
- If it's just for the purposes of importing, then it can be thought of as a kind of "view" that Paratext needs to avoid import issues while the true model is kept unmodified in handler._tokens. This allows for the option to access the unmodified usfm if needed in the future.

Pros for putting it in `process_token()`:
- Faster execution time since it's all part of the same iteration
- If thought of as an essential change to the usfm structure such that alternative views are unnecessary, it could make more structural sense to include it here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)


ddaspit (commented):
> @ddaspit reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93 and mshannon-sil).

___
*[`machine/corpora/update_usfm_parser_handler.py` line 345 at r1](https://reviewable.io/reviews/sillsdev/machine.py/285#-OoznMTk-aPDk32LvXaB:-OoznMTk-aPDk32LvXaC:b9o8fuq) ([raw file](https://github.com/sillsdev/machine.py/blob/d8ca02da8aada71b266f56c41dd14b179d76fc04/machine/corpora/update_usfm_parser_handler.py#L345)):*
> ```Python
>         tokens = list(self._tokens)
>         if chapters is not None:
>             tokens = self._get_incremental_draft_tokens(tokens, chapters)
> ```

I think we can do something similar, but before we parse instead of after. Instead of calling `parse_usfm` in `update_usfm`, we can do something like this:

```py
tokenizer = UsfmTokenizer(self._settings.stylesheet)
tokens = tokenizer.tokenize(usfm)
tokens = filter_tokens_by_chapter(tokens, chapters)
parser = UsfmParser(tokens, handler, self._settings.stylesheet, self._settings.versification)
parser.process_tokens()
```
This would avoid updating the whole book.

mshannon-sil:
*[`machine/corpora/update_usfm_parser_handler.py` line 345 at r1](https://reviewable.io/reviews/sillsdev/machine.py/285#-OoznMTk-aPDk32LvXaB:-OqHU4GM8rIan-WzEjNR:b-et0wj4) ([raw file](https://github.com/sillsdev/machine.py/blob/d8ca02da8aada71b266f56c41dd14b179d76fc04/machine/corpora/update_usfm_parser_handler.py#L345)):*

I think we can do something similar, but before we parse instead of after. Instead of calling `parse_usfm` in `update_usfm`, we can do something like this:

```py
tokenizer = UsfmTokenizer(self._settings.stylesheet)
tokens = tokenizer.tokenize(usfm)
tokens = filter_tokens_by_chapter(tokens, chapters)
parser = UsfmParser(tokens, handler, self._settings.stylesheet, self._settings.versification)
parser.process_tokens()
```
This would avoid updating the whole book.
</blockquote></details>

I updated it accordingly, how does it look now?

If we change parse_usfm to accept a Sequence[UsfmToken] as well as str, we could call it here and let it instantiate the parser and process the tokens to avoid some code duplication. But not sure if there's a reason parse_usfm only accepts str.

Enkidu93 (commented):
> @Enkidu93 reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 3 unresolved discussions (waiting on ddaspit and mshannon-sil).

___
*[`machine/corpora/paratext_project_text_updater_base.py` line 97 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPCEXUAjWbHpmZV0Ny:-OqPCEXUAjWbHpmZV0Nz:bsvizkx) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/paratext_project_text_updater_base.py#L97)):*
> ```Python
>             in_id_marker = False
>         elif token.type == UsfmTokenType.CHAPTER:
>             if token.data and int(token.data) in chapters:
> ```

I think this may be safe now after some recent changes, but you may want to double check what happens with bad chapter references like `\c 1.` if you haven't already/isn't already covered by tests.
___
*[`machine/corpora/update_usfm_parser_handler.py` line 348 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPBmT_EdhUdVh-gFWa:-OqPBmT_EdhUdVh-gFWb:b-m365me) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/update_usfm_parser_handler.py#L348)):*
> ```Python
>                 remark_tokens.append(UsfmToken(UsfmTokenType.TEXT, text=remark))
>             if len(tokens) > 0:
>                 for index, token in enumerate(tokens):
> ```

Don't we want to preserve the ability to add book-level remarks? Am I reading this correctly that we'd no longer be able to do so with this change? Peter has also put in a PR related to the per-chapter remarks; are you guys coordinating on this https://github.com/sillsdev/machine/pull/408? I think if we did something like he has there, we would have a bit more flexibility.

mshannon-sil:
*[`machine/corpora/update_usfm_parser_handler.py` line 348 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPBmT_EdhUdVh-gFWa:-OqR6-OZ-iyeSUm25pxR:bmocnvj) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/update_usfm_parser_handler.py#L348)):*

Don't we want to preserve the ability to add book-level remarks? Am I reading this correctly that we'd no longer be able to do so with this change? Peter has also put in a PR related to the per-chapter remarks; are you guys coordinating on this https://github.com/sillsdev/machine/pull/408? I think if we did something like he has there, we would have a bit more flexibility.
</blockquote></details>

I was unaware that Peter was making any changes. As far as I'm aware we want it working in SILNLP first (through machine.py) before changes to machine.

What I remember from the NT meeting where we discussed remarks was that the team wanted to move all the remarks to the chapter level once we realized that \rem is valid in chapters, so we would no longer have book remarks. I can throw the question to the wider group to see what we think.

ddaspit (commented):
> @ddaspit reviewed 4 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 4 unresolved discussions (waiting on Enkidu93 and mshannon-sil).

___
*[`machine/corpora/update_usfm_parser_handler.py` line 345 at r1](https://reviewable.io/reviews/sillsdev/machine.py/285#-OoznMTk-aPDk32LvXaB:-OqS4njX35qHjKcbolGM:b-j81t5u) ([raw file](https://github.com/sillsdev/machine.py/blob/d8ca02da8aada71b266f56c41dd14b179d76fc04/machine/corpora/update_usfm_parser_handler.py#L345)):*

I updated it accordingly, how does it look now?

If we change parse_usfm to accept a Sequence[UsfmToken] as well as str, we could call it here and let it instantiate the parser and process the tokens to avoid some code duplication. But not sure if there's a reason parse_usfm only accepts str.
</blockquote></details>

I think it makes sense to update `parse_usfm` to accept tokens as well.
___
*[`machine/corpora/update_usfm_parser_handler.py` line 348 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPBmT_EdhUdVh-gFWa:-OqS2VYU9NnBLnlXGSHR:buch26r) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/update_usfm_parser_handler.py#L348)):*

I was unaware that Peter was making any changes. As far as I'm aware we want it working in SILNLP first (through machine.py) before changes to machine.

What I remember from the NT meeting where we discussed remarks was that the team wanted to move all the remarks to the chapter level once we realized that \rem is valid in chapters, so we would no longer have book remarks. I can throw the question to the wider group to see what we think.
</blockquote></details>

The philosophy for implementing features in the Machine library is a bit different than our other applications. For applications, we only build what we need. Because Machine is a shared library, we try to keep any features we add generally useful (within reason). In this case, I think it makes sense to include the ability to add book-level remarks even if we aren't planning on using that specific feature right now.
___
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-OqS4f7aA8gJeEWQj1ly:bu1qt98) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*
> ```Python
> from .paratext_project_settings_parser_base import ParatextProjectSettingsParserBase
> from .paratext_project_terms_parser_base import KeyTerm, ParatextProjectTermsParserBase
> from .paratext_project_text_updater_base import ParatextProjectTextUpdaterBase, filter_tokens_by_chapter
> ```

This seems unnecessary.

pmachapman:
> What I remember from the NT meeting where we discussed remarks was that the team wanted to move all the remarks to the chapter level once we realized that \rem is valid in chapters, so we would no longer have book remarks. I can throw the question to the wider group to see what we think.

Sorry for jumping the gun a bit in C# land - I wanted to get something working so I could see how it might work in Serval/SF, and figure out any problems and pitfalls at this early stage.

I think book level remarks are still necessary, as we don't want to change how full book drafting is working. It could be a bit counterintuitive for users of full book drafting to suddenly have per-chapter remarks for a when they make their next draft of say Acts or Genesis, and then they need to track down and remove manually the remarks when they begin the process of editing the draft.

Also, I think we might need different remarks per chapter? Your implementation looks like it has the same remark for each draft. The only reason I thought this is that our current book level remarks in Serval:

```
\id MRK
\rem This draft of MRK was generated using AI on 2025-12-10 00:21:48Z. It should be reviewed and edited carefully.
```

Might change to per-chapter:

```
\c 3
\rem This draft of MRK 3 was generated using AI on 2025-12-10 00:21:48Z. It should be reviewed and edited carefully.
```

I think having the chapter number in the remark helps keep explicit what was drafted. Without some form of identifier in the remark, there is a danger it perhaps starts floating into another chapter if copy-pasted, or if a merge of incoming changes in PT goes haywire. I also think saying just "This draft was generated" or "The draft of this chapter was generated" is not as clear as stating exactly what chapter is drafted, and removes any ambiguity.

Apologies if I misunderstood things, or if I am wrong - I am happy which ever way we go RE: the points above - I just wanted to explain my thoughts, so you can correct me if I am barking up the wrong tree.

Enkidu93:
I put a comment in the other PR, but I think we need Matthew's updates to support partial-book USFM but I also think we should do something more like what Peter has to preserve the per-book remark functionality as well as have more fine-grained control of the per-chapter remarks. I think you guys should decide where to put the changes first (machine or machine.py) and then port.

mshannon-sil:
I've added Peter's changes for the remarks to this PR and updated test cases to reflect this. The only change I made to the approach was to add `"rem"` for `markers_to_skip` at the chapter level as well, so that we always put the remark after any existing remarks just like we do at the book level.

mshannon-sil:
*[`machine/corpora/paratext_project_text_updater_base.py` line 97 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPCEXUAjWbHpmZV0Ny:-OrG3qLO-3MCRiBvLEw0:b-qu03et) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/paratext_project_text_updater_base.py#L97)):*

I think this may be safe now after some recent changes, but you may want to double check what happens with bad chapter references like `\c 1.` if you haven't already/isn't already covered by tests.
</blockquote></details>

I updated it to use the parse_integer method to handle it more safely, and I added a test case to cover this. As is, the code now just skips over chapters with malformed chapter numbers and does not include them in the output, which you can see in the `test_filter_chapters_with_bad_chapter_reference` test case. Is this the best approach in your opinion, or do you want it to be able to select even chapters with malformed chapter numbers?

mshannon-sil:
*[`machine/corpora/update_usfm_parser_handler.py` line 345 at r1](https://reviewable.io/reviews/sillsdev/machine.py/285#-OoznMTk-aPDk32LvXaB:-OrG7CgT9i4jkSQCsLmC:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/d8ca02da8aada71b266f56c41dd14b179d76fc04/machine/corpora/update_usfm_parser_handler.py#L345)):*

I think it makes sense to update `parse_usfm` to accept tokens as well.
</blockquote></details>

Done.

mshannon-sil:
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-OrG7L-BCEyXYadqt31m:b3z18yw) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*

This seems unnecessary.
</blockquote></details>

The reason it's here is so I can call it in `test_update_usfm_parser_handler.py` since the test file reimplements `update_usfm` so that it's set up more conveniently for testing. Should we change either the test file or implementation to avoid needing to import `filter_tokens_by_chapter`?

mshannon-sil:
*[`machine/corpora/update_usfm_parser_handler.py` line 348 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPBmT_EdhUdVh-gFWa:-OrG7H2-934lsTanKEC5:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/update_usfm_parser_handler.py#L348)):*

The philosophy for implementing features in the Machine library is a bit different than our other applications. For applications, we only build what we need. Because Machine is a shared library, we try to keep any features we add generally useful (within reason). In this case, I think it makes sense to include the ability to add book-level remarks even if we aren't planning on using that specific feature right now.
</blockquote></details>

Done.

Enkidu93 (commented):
This is no longer a draft PR, correct? Or is there a reason we wouldn't want to merge this soon?

> @Enkidu93 reviewed 5 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285#-:-OrHxI_7-pbxMZP5RxHt:bps31s9)* status: all files reviewed, 2 unresolved discussions (waiting on ddaspit and mshannon-sil).

___
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-OrHvIzX2432oDn5XzSx:b-npn1xe) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*

The reason it's here is so I can call it in `test_update_usfm_parser_handler.py` since the test file reimplements `update_usfm` so that it's set up more conveniently for testing. Should we change either the test file or implementation to avoid needing to import `filter_tokens_by_chapter`?
</blockquote></details>

I don't think it has to happen now, but we should consider making a `MemoryParatextProjectTextUpdater` like we have with the other `Zip...`/`File...` Paratext classes. We could then refactor these tests to all use the memory updater and avoid the custom logic in the test`update_usfm` function as well as avoid importing `filter_tokens_by_chapter`. If we don't want to refactor all the tests, we could in the meantime just replace the logic in the `else` branch of `update_usfm` - i.e.:

```
    else:
        source = source.strip().replace("\r\n", "\n") + "\r\n"
        updater = UpdateUsfmParserHandler(
        ...
```

with a call to`MemoryParatextProjectTextUpdater.update_usfm()` which would also avoid exposing `filter_tokens_by_chapter`. I leave it to Damien whether a change like this should be done in this PR or separately later.
___
*[`machine/corpora/paratext_project_text_updater_base.py` line 97 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqPCEXUAjWbHpmZV0Ny:-OrHrKph-s3Pmw4zwzcb:bcmd4va) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/paratext_project_text_updater_base.py#L97)):*

I updated it to use the parse_integer method to handle it more safely, and I added a test case to cover this. As is, the code now just skips over chapters with malformed chapter numbers and does not include them in the output, which you can see in the `test_filter_chapters_with_bad_chapter_reference` test case. Is this the best approach in your opinion, or do you want it to be able to select even chapters with malformed chapter numbers?
</blockquote></details>

Great! I think skipping it is good for now. If, down the road, we add logic to handle these chapters more robustly elsewhere, we can add support here as well.

Enkidu93 (commented):
> @Enkidu93 resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 1 unresolved discussion (waiting on ddaspit).

mshannon-sil:
Just removed the draft designation!

ddaspit (changes requested):
> @ddaspit reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93 and mshannon-sil).

___
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-OrKODlG8hIPj9tJ4pT1:bvdjj80) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*

I don't think it has to happen now, but we should consider making a `MemoryParatextProjectTextUpdater` like we have with the other `Zip...`/`File...` Paratext classes. We could then refactor these tests to all use the memory updater and avoid the custom logic in the test`update_usfm` function as well as avoid importing `filter_tokens_by_chapter`. If we don't want to refactor all the tests, we could in the meantime just replace the logic in the `else` branch of `update_usfm` - i.e.:

```
    else:
        source = source.strip().replace("\r\n", "\n") + "\r\n"
        updater = UpdateUsfmParserHandler(
        ...
```

with a call to`MemoryParatextProjectTextUpdater.update_usfm()` which would also avoid exposing `filter_tokens_by_chapter`. I leave it to Damien whether a change like this should be done in this PR or separately later.
</blockquote></details>

I see why you are exporting it. I agree with Eli. If we want to test the chapter filtering, then we should test the `ParatextProjectTextUpdaterBase` class. The best way to do that is to create a memory-based implementation.

mshannon-sil:
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-OrKSQB3Aaki1AQxA28K:b-6glznd) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*

I see why you are exporting it. I agree with Eli. If we want to test the chapter filtering, then we should test the `ParatextProjectTextUpdaterBase` class. The best way to do that is to create a memory-based implementation.
</blockquote></details>

Sounds good, I'll work on adding that!

mshannon-sil:
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-Os4WEqeBjaqfElNNoj6:brvke8v) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*

Sounds good, I'll work on adding that!
</blockquote></details>

Added the memory implementation and verified all tests pass. I went with the option of replacing the else branch of update_usfm as suggested.

ddaspit (commented):
> @ddaspit partially reviewed 11 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 3 unresolved discussions (waiting on mshannon-sil).

___
*[`machine/corpora/__init__.py` line 30 at r4](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os6mrPJEADf17TBvyeq:-Os6mrPJEADf17TBvyer:bn8kobg) ([raw file](https://github.com/sillsdev/machine.py/blob/ef9041f9de643ed8b33b4d01e60624ce39fd8a86/machine/corpora/__init__.py#L30)):*
> ```Python
> from .paratext_project_settings_parser_base import ParatextProjectSettingsParserBase
> from .paratext_project_terms_parser_base import KeyTerm, ParatextProjectTermsParserBase
> from .paratext_project_text_updater_base import ParatextProjectTextUpdaterBase, filter_tokens_by_chapter
> ```

We don't need to export `filter_tokens_by_chapter` anymore.
___
*[`machine/corpora/memory_paratext_project_file_handler.py` line 8 at r7](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os6mX8b-Q4vEsitaujJ:-Os6mX8b-Q4vEsitaujK:b-9r9spp) ([raw file](https://github.com/sillsdev/machine.py/blob/e7f3c2cf7d79424dfaf69ddb5aa69205f68c9222/machine/corpora/memory_paratext_project_file_handler.py#L8)):*
> ```Python
> 
> 
> class MemoryParatextProjectFileHandler(ParatextProjectFileHandler):
> ```

This should stay in `tests/testutils`.
___
*[`machine/corpora/memory_paratext_project_text_updater.py` line 8 at r7](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os6m_kw-DN_7DwMCs7_:-Os6m_kw-DN_7DwMCs7a:b41gyz8) ([raw file](https://github.com/sillsdev/machine.py/blob/e7f3c2cf7d79424dfaf69ddb5aa69205f68c9222/machine/corpora/memory_paratext_project_text_updater.py#L8)):*
> ```Python
> 
> 
> class MemoryParatextProjectTextUpdater(ParatextProjectTextUpdaterBase):
> ```

This should be moved over to `tests/testutils` along with the other memory-based implementations.

mshannon-sil (commented):
> @mshannon-sil made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: 3 of 14 files reviewed, 3 unresolved discussions (waiting on ddaspit).

___
*[`machine/corpora/__init__.py` line 30 at r4](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os6mrPJEADf17TBvyeq:-Os7Gh1_7_uB3j0R3BEp:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/ef9041f9de643ed8b33b4d01e60624ce39fd8a86/machine/corpora/__init__.py#L30)):*

We don't need to export `filter_tokens_by_chapter` anymore.
</blockquote></details>

Done.
___
*[`machine/corpora/memory_paratext_project_file_handler.py` line 8 at r7](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os6mX8b-Q4vEsitaujJ:-Os7GiWW-4OsV5K8TFRe:b-896fix) ([raw file](https://github.com/sillsdev/machine.py/blob/e7f3c2cf7d79424dfaf69ddb5aa69205f68c9222/machine/corpora/memory_paratext_project_file_handler.py#L8)):*

This should stay in `tests/testutils`.
</blockquote></details>

Done.
___
*[`machine/corpora/memory_paratext_project_text_updater.py` line 8 at r7](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os6m_kw-DN_7DwMCs7_:-Os7GjF_0xNkI4LJCPeH:b-eontrz) ([raw file](https://github.com/sillsdev/machine.py/blob/e7f3c2cf7d79424dfaf69ddb5aa69205f68c9222/machine/corpora/memory_paratext_project_text_updater.py#L8)):*

This should be moved over to `tests/testutils` along with the other memory-based implementations.
</blockquote></details>

Done. I see that `memory_stream_container.py` and `memory_text.py` are in machine/corpora because other classes in machine/corpora make use of them, so I'll leave those where they are.

ddaspit (approved):
:lgtm:

> @ddaspit partially reviewed 11 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285#-:-Os7QCM34Vb765zX9nGB:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

Enkidu93 (approved):
> @Enkidu93 partially reviewed 12 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285)* status: all files reviewed, 1 unresolved discussion (waiting on mshannon-sil).

___
*[`machine/corpora/__init__.py` line 30 at r2](https://reviewable.io/reviews/sillsdev/machine.py/285#-OqS4f7aA8gJeEWQj1lx:-Os8TIg03cE0jXdLAoxv:bl4aons) ([raw file](https://github.com/sillsdev/machine.py/blob/e1865ea7afd4405b87e78b7ef1bc78df0246511c/machine/corpora/__init__.py#L30)):*

Added the memory implementation and verified all tests pass. I went with the option of replacing the else branch of update_usfm as suggested.
</blockquote></details>

Looks good. Thanks, Matthew!
___
*[`tests/testutils/memory_paratext_project_file_handler.py` line 26 at r8](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os8SPrn-QkznSryP4vc:-Os8SPrn-QkznSryP4vd:b-d6qhm4) ([raw file](https://github.com/sillsdev/machine.py/blob/a11fa51acfca8a612e0f8b1fabb5f1e888f09b71/tests/testutils/memory_paratext_project_file_handler.py#L26)):*
> ```Python
> 
>     def create_stylesheet(self, file_name: str) -> UsfmStylesheet:
>         return UsfmStylesheet(file_name)
> ```

Just for consistency's sake, you could make this mirror Machine's:
```
    public UsfmStylesheet CreateStylesheet(string fileName)
    {
        if (fileName is "usfm.sty" or "usfm_sb.sty")
            return new UsfmStylesheet(fileName);
        throw new NotImplementedException();
    }
```
But if you don't get to it, that's fine. We need to port those changes to machine.py anyways.

mshannon-sil:
*[`tests/testutils/memory_paratext_project_file_handler.py` line 26 at r8](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os8SPrn-QkznSryP4vc:-Os8WxEM7QrMVDSlqQDN:b-9qnpl7) ([raw file](https://github.com/sillsdev/machine.py/blob/a11fa51acfca8a612e0f8b1fabb5f1e888f09b71/tests/testutils/memory_paratext_project_file_handler.py#L26)):*

Just for consistency's sake, you could make this mirror Machine's:
```
    public UsfmStylesheet CreateStylesheet(string fileName)
    {
        if (fileName is "usfm.sty" or "usfm_sb.sty")
            return new UsfmStylesheet(fileName);
        throw new NotImplementedException();
    }
```
But if you don't get to it, that's fine. We need to port those changes to machine.py anyways.
</blockquote></details>

It's a quick change so I added it!

Enkidu93 (approved):
:lgtm:

> @Enkidu93 partially reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/285#-:-Os8XTML16_JcYDKdwqE:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

___
*[`tests/testutils/memory_paratext_project_file_handler.py` line 26 at r8](https://reviewable.io/reviews/sillsdev/machine.py/285#-Os8SPrn-QkznSryP4vc:-Os8XQcL6vE4fqImB4Tw:b-9755fn) ([raw file](https://github.com/sillsdev/machine.py/blob/a11fa51acfca8a612e0f8b1fabb5f1e888f09b71/tests/testutils/memory_paratext_project_file_handler.py#L26)):*

It's a quick change so I added it!
</blockquote></details>

Thank you!

PR #350, opened 2026-08-17
add low book confidence detector

### LINKED ISSUE #349: Add a function to detect low book confidence
A function should be added that will take in a book's confidence and return True if the book confidence is below a certain threshold. This should be added to `__init__.py` so it can be accessed by SILNLP. This should be ported over to the C# Machine library as well, so that it can be called by Serval.

mshannon-sil:
This adds a function to detect low book confidence. It will be used in SILNLP to flag books with low confidence, and it will be ported to Machine so Serval can use it. It's quite straightforward, so let me know if there's any additional logic we should be adding to machine.py.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/350)

### DIFF
--- a/machine/quality_estimation/__init__.py
+++ b/machine/quality_estimation/__init__.py
@@ -1,3 +1,4 @@
+from .book_confidence import LOW_BOOK_CONFIDENCE_THRESHOLD, is_book_confidence_unusually_low
 from .chrf3_quality_estimator import ChrF3QualityEstimator
 from .scripture_book_usability import ScriptureBookUsability
 from .scripture_chapter_usability import ScriptureChapterUsability
@@ -11,6 +12,8 @@ from .usability_parameters import UsabilityParameters
 
 __all__ = [
     "ChrF3QualityEstimator",
+    "is_book_confidence_unusually_low",
+    "LOW_BOOK_CONFIDENCE_THRESHOLD",
     "ScriptureBookUsability",
     "ScriptureChapterUsability",
     "ScriptureSegmentUsability",
--- a/machine/quality_estimation/book_confidence.py
+++ b/machine/quality_estimation/book_confidence.py
@@ -1,7 +1,11 @@
+from typing import Optional
+
 LOW_BOOK_CONFIDENCE_THRESHOLD = 0.42
 
 
-def is_book_confidence_unusually_low(confidence: float) -> bool:
+def is_book_confidence_unusually_low(
+    confidence: float, book_id: Optional[str] = None, model: Optional[str] = None
+) -> bool:
     if not 0 <= confidence <= 1:
         raise ValueError(
             f"The book confidence {confidence} is invalid. "
--- a/machine/quality_estimation/book_confidence.py
+++ b/machine/quality_estimation/book_confidence.py
@@ -1,0 +1,11 @@
+LOW_BOOK_CONFIDENCE_THRESHOLD = 0.42
+
+
+def is_book_confidence_unusually_low(confidence: float) -> bool:
+    if not 0 <= confidence <= 1:
+        raise ValueError(
+            f"The book confidence {confidence} is invalid. "
+            f"It is calculated as the geometric mean of the segment confidences, "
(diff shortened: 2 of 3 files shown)


pmachapman (approved):
:lgtm: Just one question!

> @pmachapman reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/350#-:-P-Gbf57D6vhW5A9KwRs:b2sxyl8)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93 and mshannon-sil).

___
*[`tests/quality_estimation/test_book_confidence.py` line 30 at r1](https://reviewable.io/reviews/sillsdev/machine.py/350#-P-GcG0TDZ9tpFgqaTXf:-P-GcG0TDZ9tpFgqaTXg:b2lnnkz) ([raw file](https://github.com/sillsdev/machine.py/blob/d6ccce134a567fcad446ddf0fa0b61a3766a27ae/tests/quality_estimation/test_book_confidence.py#L30)):*
> ```Python
> def test_is_book_confidence_unusually_low_nan() -> None:
>     with raises(ValueError):
>         is_book_confidence_unusually_low(float("nan"))
> ```

Right now `NmtEngineBuildJob._train_model()` returns a confidence value of `float("nan")`. Do we want to handle this as a special not defined case (and so ignore), or do you think it should be an error case as per your implementation?



_Code quote:_
```Python
def test_is_book_confidence_unusually_low_nan() -> None:
    with raises(ValueError):
        is_book_confidence_unusually_low(float("nan"))
```

mshannon-sil:
*[`tests/quality_estimation/test_book_confidence.py` line 30 at r1](https://reviewable.io/reviews/sillsdev/machine.py/350#-P-GcG0TDZ9tpFgqaTXf:-P-Gkvp79fVpVcrdLjcx:b-xqlwf8) ([raw file](https://github.com/sillsdev/machine.py/blob/d6ccce134a567fcad446ddf0fa0b61a3766a27ae/tests/quality_estimation/test_book_confidence.py#L30)):*
<details><summary><i>Previously, pmachapman (Peter Chapman) wrote…</i></summary><blockquote>

Right now `NmtEngineBuildJob._train_model()` returns a confidence value of `float("nan")`. Do we want to handle this as a special not defined case (and so ignore), or do you think it should be an error case as per your implementation?
</blockquote></details>

Would this function ever see the output of `NmtEngineBuildJob._train_model()`, seeing as how this function is only meant to run on book confidence scores?

I'm good with doing whatever is most consistent with how the machine/machine.py repo has been handling nan. @Enkidu93 any thoughts?

mshannon-sil:
I added one more change so that the method matches the current function stub's signature in Serval.

Enkidu93 (approved):
:lgtm:

> @Enkidu93 partially reviewed 3 files and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/350#-:-P-GqCvz88XGCxHBQ4ZR:bnfp4nl)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on mshannon-sil and pmachapman).

___
*[`tests/quality_estimation/test_book_confidence.py` line 30 at r1](https://reviewable.io/reviews/sillsdev/machine.py/350#-P-GcG0TDZ9tpFgqaTXf:-P-GpV1e8erl5xH-VoK8:bhk5jh) ([raw file](https://github.com/sillsdev/machine.py/blob/d6ccce134a567fcad446ddf0fa0b61a3766a27ae/tests/quality_estimation/test_book_confidence.py#L30)):*

Would this function ever see the output of `NmtEngineBuildJob._train_model()`, seeing as how this function is only meant to run on book confidence scores?

I'm good with doing whatever is most consistent with how the machine/machine.py repo has been handling nan. @Enkidu93 any thoughts?
</blockquote></details>

Yeah, I don't think we need to worry about this since the confidence coming from `_train_model` is a different sort of confidence value relevant for SMT.

PR #363, opened 2026-09-10
Migrate to transformers 5

pmachapman:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363)


Enkidu93:
Adding @mshannon-sil as a reviewer here as well. Thank you for doing this, Peter!

Enkidu93 (approved):
:lgtm:

I know this was a lot of work. Thank you, Peter! These kinds of changes always make me nervous. Unfortunately, Mike is no longer doing his BLEU tests (he mentioned SF is picking those up somehow? Not sure what he meant exactly?). Maybe we should begin doing that sort of test ourselves. 

> @Enkidu93 reviewed 17 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363#-:-P1WOuHJEonaW6IgY--c:b-40b97t)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ddaspit and mshannon-sil).

ddaspit (commented):
Thanks for taking this on. A few things I found while going through it; the `batch_prepare_for_model` one is the only one I think has to be fixed before merging.

One more that isn't in the diff: `samples/machine_translation.ipynb` still passes `overwrite_output_dir=True`, which is gone in v5, so that cell will raise.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
```diff
@@ -304,37 +296,23 @@ def batch_prepare_for_model(
...
-                outputs = tokenizer.prepare_for_model(ids, add_special_tokens=False)
-
-                for key, value in outputs.items():
-                    if key not in batch_outputs:
-                        batch_outputs[key] = []
-                    batch_outputs[key].append(value)
-            return BatchEncoding(batch_outputs, tensor_type=return_tensors)
+            return tokenizer(
+                batch_tokens,
+                is_split_into_words=True,
```
ddaspit, line 301: `is_split_into_words=True` re-runs the pre-tokenizer on each piece, so continuation pieces get a `▁` prepended and get re-split. I checked this against nllb-200-distilled-600M: `['▁tele', 'phones']` comes back as `['▁tele', '▁ph', 'ones']`, and the gloss pieces `['tele', 'phone']` come back as `['▁tele', '▁phone']`, which undoes the `▁` stripping above. Any batch with a key-term row trains on the wrong ids.

Can we go back to `convert_tokens_to_ids` and build the `input_ids`/`attention_mask` dict by hand? That still works in v5.

machine/translation/huggingface/hugging_face_nmt_engine.py
```diff
@@ -309,15 +329,26 @@ def _forward(self, model_inputs, **generate_kwargs):
         start_index = 0
         if self.model.config.decoder_start_token_id is not None:
             start_index = 1
-        if generate_kwargs["output_attentions"] is True:
+        if self.generation_config.output_attentions:
```
ddaspit, line 332: The old code defaulted `output_attentions` to True here. `self.generation_config.output_attentions` defaults to None in v5, so anyone who doesn't pass the flag now gets empty alignments. Meanwhile line 53 still picks eager attention when the flag is missing, so we pay for eager and get nothing from it. Should the two agree on a default?

machine/translation/huggingface/hugging_face_nmt_engine.py
```diff
@@ -14,47 +15,52 @@
...
+        if isinstance(model, PreTrainedModel):
+            self._model = model
             self._model.eval()
             self._is_model_owned = False
         else:
-            model_config = AutoConfig.from_pretrained(str(self._model), label2id={}, id2label={}, num_labels=0)
+            model_config = AutoConfig.from_pretrained(str(model), label2id={}, id2label={}, num_labels=0)
+
+            # If output_attentions is True or None, we need to set the attn_implementation to eager to get the attentions
+            attn_implementation = "eager" if self._pipeline_kwargs.get("output_attentions", True) else "sdpa"
```
ddaspit, line 53: This only sets eager when loading from a path. If someone passes a model object (which `HuggingFaceNmtModel` does) with `output_attentions=True`, the model stays on sdpa, sdpa returns None for the weights, and `attentions[0][0]` blows up. v4 fell back to eager here; v5 doesn't. Maybe call `set_attn_implementation("eager")` on that branch too.

machine/translation/huggingface/transformers_compatibility.py
```diff
@@ -0,0 +1,151 @@
...
+
+    _pipeline_calls_generate = True
+    _load_processor = False
+    _load_image_processor = False
+    _load_feature_extractor = False
+    _load_tokenizer = True
+    # Make sure the docstring is updated when the default generation config is changed (in all pipelines in this file)
+    _default_generation_config = GenerationConfig(
+        max_new_tokens=256,
+        num_beams=4,
```
ddaspit, line 26: `num_beams=4` here overrides the model's own config because the pipeline merge only fills in None values. So an engine built without `num_beams` now does 4-beam search. That's why `test_translate_greedy` needed new expected output and a real sequence confidence: it isn't greedy anymore. I'd drop `num_beams` from this default and put the greedy test back.

machine/jobs/settings.yaml
```diff
@@ -7,13 +7,14 @@ default:
   huggingface:
     parent_model_name: facebook/nllb-200-distilled-1.3B
     train_params:
+      auto_find_batch_size: true
```
ddaspit, line 10: The trainer we removed doubled `gradient_accumulation_steps` every time it halved the batch, so the effective batch stayed at 64. The HF version just shrinks the batch (by 0.9x per retry in accelerate 1.14, so it can take six or more retries to get from 64 to 32) and leaves accumulation alone. With `max_steps: 5000` that means fewer examples seen and a warmup/cosine schedule tuned for a batch we're no longer using. Is that intended? If so, a comment here would help.

Also, `TrainParams` doesn't expose `auto_find_batch_size`, so clients can't turn it off.

tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py
```diff
@@ -185,14 +182,14 @@ def test_update_tokenizer_missing_char() -> None:
...
-            norm_result_char2 = finetuned_engine_char.tokenizer.backend_tokenizer.normalizer.normalize_str("‍")
+            norm_result_char1 = finetuned_engine_char.tokenizer.encode("‌ ")
+            norm_result_char2 = finetuned_engine_char.tokenizer.encode("‍")
 
         assert norm_result_nochar1 != norm_result_char1
         assert norm_result_nochar2 != norm_result_char2
 
         assert finetuned_result_nochar != finetuned_result_char
-        assert finetuned_result_nochar_composite != finetuned_result_char_composite
+        assert finetuned_result_nochar_composite == finetuned_result_char_composite
```
ddaspit, line 192: This changed from `!=` to `==`, so adding missing characters no longer changes how `Ḏ` is tokenized. Do we know why? I couldn't find an id or normalizer problem on reload, so I don't have an explanation. It'd be good to have one before locking in the new expectation.

machine/jobs/nmt_build_options.py
```diff
@@ -15,7 +15,7 @@ class TrainParams(BaseModel):
     per_device_train_batch_size: int | None = None
     gradient_accumulation_steps: int | None = None
     label_smoothing_factor: float | None = None
-    group_by_length: bool | None = None
+    train_sampling_strategy: str | None = None
```
ddaspit, line 18: `TrainParams` has `extra="forbid"`, so any client still sending `group_by_length` gets a validation error. Can we accept both names for a while, or at least call this out as a breaking change?

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
```diff
@@ -125,17 +112,21 @@ def train(
         check_canceled: Optional[Callable[[], None]] = None,
     ) -> None:
         last_checkpoint = None
-        if os.path.isdir(self._training_args.output_dir) and not self._training_args.overwrite_output_dir:
+        if (
+            self._training_args.output_dir is not None
+            and os.path.isdir(self._training_args.output_dir)
+            and self._training_args.resume_from_checkpoint is not None
```
ddaspit, line 118: Now that the outer `if` requires `resume_from_checkpoint is not None`, the inner `elif ... is None` can't run, and the `elif last_checkpoint` further down can't either. Auto-resume from a checkpoint in `output_dir` is effectively gone. Also, if someone passes an explicit checkpoint path and `output_dir` already has a saved model in it, this raises "already exists and is not empty" with no way around it.

I think the whole block can go and we just pass `resume_from_checkpoint` straight to `trainer.train()`; the Trainer already handles both cases.

machine/translation/huggingface/hugging_face_nmt_engine.py
```diff
@@ -272,15 +278,29 @@ def _forward(self, model_inputs, **generate_kwargs):
...
-        transition_scores = cast(
-            torch.Tensor,
-            self.model.compute_transition_scores(
-                output_ids,  # type: ignore
-                scores,  # type: ignore
-                beam_indices,  # type: ignore
-                normalize_logits=True,
-            ),
-        )
+        try:
```
ddaspit, line 281: What exception was this added for? `except Exception` catches CUDA OOM too, which means the batch-size backoff in `translate_n_batch` never sees it and we quietly redo the whole thing on CPU every call. The fallback also moves `output_ids` but not `attentions`, so with attentions on it'd fail with a device mismatch. I'd go back to the single call.

Separately, beam scores are already log-probs in v5, so `normalize_logits=False` would cut the biggest allocation here.

machine/translation/huggingface/transformers_compatibility.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __init__(
+        self,
+        framework: Optional[str] = "pt",
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.framework = framework
+
+    def _sanitize_parameters(
```
ddaspit, line 37: `ReturnType`, `return_tensors`, `return_text`, `return_type`, `stop_sequence`, and `framework` aren't used by anything. Passing `return_tensors` would actually error in `postprocess` since it has no `return_type` param. Fine to trim these down to what we need (`src_lang`, `tgt_lang`, `truncation`, `_parse_and_tokenize`, `preprocess`).

machine/translation/huggingface/hugging_face_nmt_engine.py
```diff
@@ -236,33 +243,32 @@ def preprocess(self, *args, truncation=TruncationStrategy.DO_NOT_TRUNCATE, src_l
...
+        if "input_tokens" in model_inputs:
+            input_tokens = model_inputs.pop("input_tokens")
         else:
-            config = self.model.config
-        generate_kwargs["min_length"] = generate_kwargs.get("min_length", config.min_length)
-        generate_kwargs["max_length"] = generate_kwargs.get("max_length", config.max_length)
-        generate_kwargs["output_attentions"] = generate_kwargs.get("output_attentions", True)
-        self.check_inputs(input_length, generate_kwargs["min_length"], generate_kwargs["max_length"])
-        output = self.model.generate(
+            input_tokens = [self.tokenizer.convert_ids_to_tokens(seq) for seq in model_inputs["input_ids"]]
```
ddaspit, line 253: `preprocess` always sets `input_tokens`, so this fallback can't run. If it ever did, it'd include special tokens and disagree with the filtering in `postprocess`. Suggest just `model_inputs.pop("input_tokens")` like before.

machine/translation/huggingface/hugging_face_nmt_engine.py
```diff
@@ -84,24 +91,24 @@ def __init__(
...
 
         self._batch_size = int(self._pipeline_kwargs.pop("batch_size", 1))
 
         self._oom_batch_size_backoff_mult = oom_batch_size_backoff_mult
 
-        self._pipeline = _TranslationPipeline(
+        self._pipeline = SilTranslationPipeline(
             model=self._model,
-            tokenizer=self._tokenizer,
+            tokenizer=cast(PreTrainedTokenizer, self.tokenizer),
```
ddaspit, line 111: This casts to `PreTrainedTokenizer` but line 151 passes the raw tokenizer. In v5 the fast tokenizer isn't a subclass of `PreTrainedTokenizer`, so the cast is lying. Suggest typing the pipeline param as `PreTrainedTokenizerBase` and passing `self._tokenizer` in both places.

machine/translation/huggingface/hugging_face_nmt_model.py
```diff
@@ -18,14 +17,14 @@
...
-        model: Union[PreTrainedModel, StrPath],
+        model: Union[PreTrainedModel, Path],
         parent_model_name: str,
         training_args: Optional[Seq2SeqTrainingArguments] = None,
         **pipeline_kwargs,
     ) -> None:
         self._model = model
         if isinstance(model, PreTrainedModel):
-            self._model_path = Path(model.name_or_path)
+            self._model_path = Path(str(model.name_or_path))
```
ddaspit, line 27: `Path(str(None))` turns a missing `name_or_path` into a directory called "None". Better to raise here. Also, dropping `StrPath` means this no longer accepts `str` under pyright while everything else in the package does.

machine/jobs/settings.yaml
```diff
@@ -7,13 +7,14 @@ default:
...
       do_train: true
       optim: adamw_torch
       warmup_steps: 1000
       per_device_train_batch_size: 64
       gradient_accumulation_steps: 1
       label_smoothing_factor: 0.2
-      group_by_length: true
+      train_sampling_strategy : group_by_length
```
ddaspit, line 17: Nit: stray space before the colon.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
```diff
@@ -176,19 +167,22 @@ def train(
         def find_missing_characters(tokenizer: Any, train_dataset: Dataset, lang_codes: List[str]) -> List[str]:
             vocab = tokenizer.get_vocab().keys()
             charset = set()
-            mpn_normalize = True if isinstance(tokenizer, (NllbTokenizerFast)) else False
+            mpn_normalize = True if isinstance(tokenizer, NllbTokenizer) else False
```
ddaspit, line 170: Nit: `True if isinstance(...) else False` is just `isinstance(...)`.

mshannon-sil (changes requested):
> @mshannon-sil reviewed 8 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363)* status: all files reviewed, 17 unresolved discussions (waiting on pmachapman).

___
*[`.gitignore` line 147 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1a7tJ5Ee6qNOpFmQdb:-P1a7tJ5Ee6qNOpFmQdc:b-i6nyy5) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/.gitignore#L147)):*
> ```ignore
> 
> # Ignore custom pyright configuration
> pyrightconfig.json
> ```

Do we not want a standard pyrightconfig.json for the repository?
___
*[`machine/translation/huggingface/transformers_compatibility.py` line 1 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1aIcIJ6RCYATURFfd3:-P1aIcIKA08ZQQFMkJJ1:bu3dpry) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/machine/translation/huggingface/transformers_compatibility.py#L1)):*
> ```Python
> import enum
> ```

What are the benefits of recreating the `TranslationPipeline` class ourselves, now that it's been removed from `transformers`? Is it just that it's more similar to what was here before? SILNLP no longer imports `Pipeline` at all from `transformers`, choosing to create a standalone`SilTranslator` class instead. I had Devin review the two approaches, and it seems that using something similar to SILNLP's approach here could cut down on overhead and reduce opportunities for bugs. It also seems to make intuitive sense to me that we shouldn't need to create a whole file for transformers compatibility to replicate transformers 4.x code, if instead we can rearchitect it to feel like more natural transformers 5.x code. But I  admit, I don't have a 100% understanding of the pros and cons of both approaches, so feel free to share if there's a production, repo, or other reason why keeping the pipeline makes more sense.

machine/jobs/nmt_build_options.py
pmachapman: Done. I've added it and updated `HuggingFaceNmtModelFactory` to migrate this property to `train_sampling_strategy` and warn the user it is deprecated.

machine/jobs/settings.yaml
pmachapman: After some experimentation, I have reverted to the previous `AutoGradientAccumulationStepsSeq2SeqTrainer`, and updated it to work correctly with the new version of transformers.

machine/jobs/settings.yaml
pmachapman: Done.

machine/translation/huggingface/hugging_face_nmt_engine.py
pmachapman: > Maybe call `set_attn_implementation("eager")` on that branch too.

Done.

machine/translation/huggingface/hugging_face_nmt_engine.py
pmachapman: Done.

machine/translation/huggingface/hugging_face_nmt_engine.py
pmachapman: Done.

machine/translation/huggingface/hugging_face_nmt_engine.py
pmachapman: This code was ported from SILNLP's SilTranslator. I've removed the `try...except`.

machine/translation/huggingface/hugging_face_nmt_engine.py
pmachapman: I've made the default `True`.

machine/translation/huggingface/hugging_face_nmt_model.py
pmachapman: I've added a `None` check, and added `str` to the allowable `model` values. `StrPath` caused pyright to throw an error.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
pmachapman: Agreed - done.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
pmachapman: Done.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
pmachapman: Done. `prepare_for_model` is removed, but `_encode_plus` is approximately the same.

machine/translation/huggingface/transformers_compatibility.py
pmachapman: I removed `num_beams` and the only value which reverted in the test was the translated string, while the confidence value went really weird - 1.39. Running the tests on a GPU were fine, but on a CPU, Math Overflows occurred.

These defaults came straight from the code in transformers v4 for `Text2TextGenerationPipeline`, so I imagine they were used? I have kept them unless you notice something else is causing the odd behavior.

machine/translation/huggingface/transformers_compatibility.py
pmachapman: Done.

tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py
pmachapman: I don't have a clear idea - the closest to a reason I can come to is the difference in normalizer. This test stumped me for a while.

pmachapman (commented):
> One more that isn't in the diff: `samples/machine_translation.ipynb` still passes `overwrite_output_dir=True`, which is gone in v5, so that cell will raise.

Done. Thanks!

> @pmachapman made 18 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363#-:-P1fp_AUDMx3XFYEFwCz:b-gbjdz2)* status: 8 of 18 files reviewed, 15 unresolved discussions (waiting on ddaspit, Enkidu93, and mshannon-sil).

___
*[`machine/translation/huggingface/transformers_compatibility.py` line 1 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1aIcIJ6RCYATURFfd3:-P1ffilH08J6G9xAymDO:bk0uhip) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/machine/translation/huggingface/transformers_compatibility.py#L1)):*

What are the benefits of recreating the `TranslationPipeline` class ourselves, now that it's been removed from `transformers`? Is it just that it's more similar to what was here before? SILNLP no longer imports `Pipeline` at all from `transformers`, choosing to create a standalone`SilTranslator` class instead. I had Devin review the two approaches, and it seems that using something similar to SILNLP's approach here could cut down on overhead and reduce opportunities for bugs. It also seems to make intuitive sense to me that we shouldn't need to create a whole file for transformers compatibility to replicate transformers 4.x code, if instead we can rearchitect it to feel like more natural transformers 5.x code. But I  admit, I don't have a 100% understanding of the pros and cons of both approaches, so feel free to share if there's a production, repo, or other reason why keeping the pipeline makes more sense.
</blockquote></details>

I needed to implement Pipeline for batch support. My initial implementation was with a copy of `SilTranslator`, but due to its lack of batch support, I could make the tests pass, but it failed when used on Serval.

My understanding of the changes in transformers 5 is just that they removed the `Text2TextGenerationPipeline`, with their suggestion being to use an LLM instead - see https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md#text-pipelines-that-should-just-be-llms. I don't think the LLM approach will work for us just yet?

My actual porting of `TranslationPipeline` wasn't too involved - most of the differences are me stripping out code that is unnecessary or unused, and combining `TranslationPipeline` and `Text2TextGenerationPipeline`.
___
*[`.gitignore` line 147 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1a7tJ5Ee6qNOpFmQdb:-P1fho312tU24iCxdjem:b6ewx) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/.gitignore#L147)):*

Do we not want a standard pyrightconfig.json for the repository?
</blockquote></details>

I've removed this line - I can't find the file.

machine/translation/huggingface/hugging_face_nmt_engine.py
ddaspit: I realized that this exception was added in silnlp to deal with an OOM error that can occur when retrieving the scores. It would fallback to computing the transition scores on the CPU, when an OOM error occurred on the GPU. I added a commit that has what I think is a better fix for this problem. It provides our own implementation of the function that retrieves just the scores we want.

machine/translation/huggingface/transformers_compatibility.py
ddaspit: Removing `num_beams` caused it to use a greedy search, which had a bug that is fixed in my commit.

tests/translation/huggingface/test_hugging_face_nmt_model_trainer.py
ddaspit: This was caused by a regression in `MbartTokenizer` in transformers 5. I implemented a workaround.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 18 files and all commit messages, made 4 comments, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363#-:-P1mWLct04dXEf-eM-dA:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on Enkidu93, mshannon-sil, and pmachapman).

mshannon-sil:
*[`machine/translation/huggingface/transformers_compatibility.py` line 1 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1aIcIJ6RCYATURFfd3:-P1r5z2L9iWWCDEywIq-:b-wu6cp4) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/machine/translation/huggingface/transformers_compatibility.py#L1)):*
<details><summary><i>Previously, pmachapman (Peter Chapman) wrote…</i></summary><blockquote>

I needed to implement Pipeline for batch support. My initial implementation was with a copy of `SilTranslator`, but due to its lack of batch support, I could make the tests pass, but it failed when used on Serval.

My understanding of the changes in transformers 5 is just that they removed the `Text2TextGenerationPipeline`, with their suggestion being to use an LLM instead - see https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md#text-pipelines-that-should-just-be-llms. I don't think the LLM approach will work for us just yet?

My actual porting of `TranslationPipeline` wasn't too involved - most of the differences are me stripping out code that is unnecessary or unused, and combining `TranslationPipeline` and `Text2TextGenerationPipeline`.
</blockquote></details>

Is there a way that Serval is calling the pipeline that is unique to Serval and not how machine.py works? I've looked through the machine.py master branch, and it seems _TranslationPipeline was only ever called on a single batch at a time, since it was called inside of `try_translate_n_batch`. If Serval works differently, is there a reason why it's not going through `translate_n_batch` to take care of the batching? Does it need different batching logic?

mshannon-sil:
*[`machine/translation/huggingface/transformers_compatibility.py` line 1 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1aIcIJ6RCYATURFfd3:-P1r9-700YAaAJznaCUR:b-ca88vh) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/machine/translation/huggingface/transformers_compatibility.py#L1)):*

Is there a way that Serval is calling the pipeline that is unique to Serval and not how machine.py works? I've looked through the machine.py master branch, and it seems _TranslationPipeline was only ever called on a single batch at a time, since it was called inside of `try_translate_n_batch`. If Serval works differently, is there a reason why it's not going through `translate_n_batch` to take care of the batching? Does it need different batching logic?
</blockquote></details>

And yes, your understanding of the changes are correct. We're also not at a spot to replace it with an LLM call.

machine/jobs/huggingface/hugging_face_nmt_model_factory.py
```diff
@@ -26,7 +27,15 @@ def __init__(self, config: Any) -> None:
...
+        # Allow group_by_length backwards compatibility
+        if "group_by_length" in args:
+            warnings.warn(
+                "'group_by_length' is deprecated and will be removed in a future release.",
+                category=DeprecationWarning,
+                stacklevel=2,
+            )
+            if args.pop("group_by_length") is True:
+                args["train_sampling_strategy"] = "group_by_length"
         # Use "max_steps" from root for backward compatibility
```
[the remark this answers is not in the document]
pmachapman: Done.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
```diff
@@ -409,7 +379,6 @@ def preprocess_function(examples):
             model=model,
             args=self._training_args,
             train_dataset=cast(Any, train_dataset),
-            tokenizer=tokenizer,
             data_collator=data_collator,
             callbacks=[
                 _ProgressCallback(
```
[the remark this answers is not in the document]
pmachapman: Done.

machine/translation/huggingface/hugging_face_nmt_model_trainer.py
```diff
@@ -500,64 +464,57 @@ def __init__(
...
-            compute_metrics,
-            callbacks,
-            optimizers,  # type: ignore
-            preprocess_logits_for_metrics,
+            model=model,
+            args=args,
+            data_collator=data_collator,
+            train_dataset=train_dataset,
+            callbacks=callbacks,
         )
```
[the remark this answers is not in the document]
pmachapman: Done.

machine/translation/huggingface/transformers_compatibility.py
```diff
@@ -0,0 +1,118 @@
...
+        tgt_lang: str | None = None,
+    ):
+        if self.tokenizer is None:
+            raise RuntimeError("No tokenizer is specified.")
+        build_inputs = getattr(self.tokenizer, "_build_translation_inputs", None)
+        if callable(build_inputs):
+            build_inputs_fn = cast(Callable[..., Any], build_inputs)
+            return build_inputs_fn(
+                *args, return_tensors="pt", truncation=truncation, src_lang=src_lang, tgt_lang=tgt_lang
+            )
```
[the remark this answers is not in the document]
pmachapman: Done.

machine/translation/huggingface/transformers_compatibility.py
```diff
@@ -0,0 +1,113 @@
...
```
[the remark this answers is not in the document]
pmachapman: Done.

pmachapman (commented):
> @pmachapman made 5 comments, resolved 6 discussions, and dismissed @mshannon-sil from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363)* status: 17 of 18 files reviewed, 1 unresolved discussion (waiting on ddaspit, Enkidu93, and mshannon-sil).

ddaspit (commented):
> @ddaspit reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/363)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93 and mshannon-sil).

___
*[`machine/translation/huggingface/transformers_compatibility.py` line 1 at r2](https://reviewable.io/reviews/sillsdev/machine.py/363#-P1aIcIJ6RCYATURFfd3:-P2Jz4Kq3vbK6zbYvSSv:b-4hlk5o) ([raw file](https://github.com/sillsdev/machine.py/blob/1c4d6417425fa6947377109082663a093e8ca346/machine/translation/huggingface/transformers_compatibility.py#L1)):*

And yes, your understanding of the changes are correct. We're also not at a spot to replace it with an LLM call.
</blockquote></details>

One of the key benefits of reimplementing `TranslationPipeline` is that it helps us to maintain backwards compatibility. Batching is an example of this. There are two levels of batching at play here. The first is at the Machine API level. This is used to support the streaming APIs in Machine. The second is at the Huggingface level. A model will often have a batch size for training/inferencing that needs to be configured to maximize speed and memory usage for a model. This is a separate batch size that can be configured for a particular translation engine. To keep Machine flexible, we want to support both types of batches. There are other configuration options that can be passed to the pipeline that we want to continue to support for compatibility reasons (see `pipeline_kwargs` argument in the `HuggingFaceNmtEngine` constructor).

PR #364, opened 2026-09-11
Fix unclosed style marker crash

benjaminking:
This PR corrects a few different errors that could occur with marker placement, due to unmatched markers in the source, one of which was causing a crash, and the others leading to incorrect output:

* An unclosed style marker in a non-verse paragraph would cause a crash
* An unclosed style marker would cause the next occurring paragraph marker to be ignored
* Only the first row of a verse range was being read
* An element of type `OTHER` could cause markers to be placed out of order
* Plus one corner case where `end_note` was behaving inconsistently in duplicated verses

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364)

### DIFF
--- a/machine/corpora/__init__.py
+++ b/machine/corpora/__init__.py
@@ -58,12 +58,8 @@ from .token_processors import (
     normalize,
     unescape_spaces,
 )
-from .update_usfm_parser_handler import (
-    UpdateUsfmMarkerBehavior,
-    UpdateUsfmParserHandler,
-    UpdateUsfmRow,
-    UpdateUsfmTextBehavior,
-)
+from .update_usfm_behavior import UpdateUsfmMarkerBehavior, UpdateUsfmTextBehavior
+from .update_usfm_parser_handler import UpdateUsfmParserHandler, UpdateUsfmRow
 from .usfm_file_text import UsfmFileText
 from .usfm_file_text_corpus import UsfmFileTextCorpus
 from .usfm_memory_text import UsfmMemoryText
--- a/machine/corpora/paratext_project_text_updater_base.py
+++ b/machine/corpora/paratext_project_text_updater_base.py
@@ -5,12 +5,8 @@ from ..utils.string_utils import parse_integer
 from .paratext_project_file_handler import ParatextProjectFileHandler
 from .paratext_project_settings import ParatextProjectSettings
 from .paratext_project_settings_parser_base import ParatextProjectSettingsParserBase
-from .update_usfm_parser_handler import (
-    UpdateUsfmMarkerBehavior,
-    UpdateUsfmParserHandler,
-    UpdateUsfmRow,
-    UpdateUsfmTextBehavior,
-)
+from .update_usfm_behavior import UpdateUsfmMarkerBehavior, UpdateUsfmTextBehavior
+from .update_usfm_parser_handler import UpdateUsfmParserHandler, UpdateUsfmRow
 from .usfm_parser import parse_usfm
 from .usfm_token import UsfmTokenType
 from .usfm_tokenizer import UsfmToken, UsfmTokenizer
--- a/machine/corpora/place_markers_usfm_update_block_handler.py
+++ b/machine/corpora/place_markers_usfm_update_block_handler.py
@@ -21,6 +21,24 @@ class PlaceMarkersAlignmentInfo(TypedDict):
     style_behavior: UpdateUsfmMarkerBehavior
 
 
+def _element_text(element: UsfmUpdateBlockElement) -> str:
+    # A TEXT element holds one token per row matched to the block, so a verse range matched by
+    # several rows has several tokens. Reading only the first drops the rest of the translation.
+    return "".join(t.to_usfm() for t in element.tokens)
+
+
+def _is_placeable(element: UsfmUpdateBlockElement, alignment_info: PlaceMarkersAlignmentInfo) -> bool:
+    # An element marked for removal is never placed, so it must not keep the block from
+    # returning early below: the placement code would then have nothing left to place.
+    if element.marked_for_removal:
+        return False
+    if element.type == UsfmUpdateBlockElementType.PARAGRAPH:
+        return alignment_info["paragraph_behavior"] == UpdateUsfmMarkerBehavior.PRESERVE and len(element.tokens) == 1
+    if element.type == UsfmUpdateBlockElementType.STYLE:
+        return alignment_info["style_behavior"] == UpdateUsfmMarkerBehavior.PRESERVE
+    return False
+
+
 class PlaceMarkersUsfmUpdateBlockHandler(UsfmUpdateBlockHandler):
     def __init__(self, *args):
         super().__init__(*args)
@@ -38,20 +56,7 @@ class PlaceMarkersUsfmUpdateBlockHandler(UsfmUpdateBlockHandler):
             len(elements) == 0
             or alignment_info["alignment"].row_count == 0
             or alignment_info["alignment"].column_count == 0
-            or not any(
-                (
-                    (
-                        e.type == UsfmUpdateBlockElementType.PARAGRAPH
-                        and alignment_info["paragraph_behavior"] == UpdateUsfmMarkerBehavior.PRESERVE
-                        and len(e.tokens) == 1
-                    )
-                    or (
-                        e.type == UsfmUpdateBlockElementType.STYLE
-                        and alignment_info["style_behavior"] == UpdateUsfmMarkerBehavior.PRESERVE
-                    )
-                )
-                for e in elements
-            )
+            or not any(_is_placeable(e, alignment_info) for e in elements)
         ):
             return block
 
@@ -74,7 +79,9 @@ class PlaceMarkersUsfmUpdateBlockHandler(UsfmUpdateBlockHandler):
                         elements.pop(i)
             elif not (
                 element.type == UsfmUpdateBlockElementType.EMBED
-                or (element.type == UsfmUpdateBlockElementType.TEXT and len(element.tokens[0].to_usfm().strip()) == 0)
+                # OTHER elements are never transferred, so they must not count as content here
+                or element.type == UsfmUpdateBlockElementType.OTHER
+                or (element.type == UsfmUpdateBlockElementType.TEXT and len(_element_text(element).strip()) == 0)
             ):
                 eob_empty_paras = False
 
@@ -92,7 +99,7 @@ class PlaceMarkersUsfmUpdateBlockHandler(UsfmUpdateBlockHandler):
         for element in elements:
             if element.type == UsfmUpdateBlockElementType.TEXT:
                 if element.marked_for_removal:
-                    text = element.tokens[0].to_usfm()
+                    text = _element_text(element)
                     src_sent += text
 
                     # Track seen tokens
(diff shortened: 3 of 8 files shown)


Enkidu93 (changes requested):
> @Enkidu93 reviewed 8 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364)* status: all files reviewed, 3 unresolved discussions (waiting on benjaminking and ddaspit).

___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 732 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UvlcOCQ4tLCJg9XyC:-P1UvlcOCQ4tLCJg9XyD:b-67c2kx) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L732)):*
> ```Python
> 
> 
> def test_unclosed_style_marker_in_non_verse_paragraph() -> None:
> ```

I'm assuming that all these tests were previously failing since you've added them or are covering a portion of uncovered code? I'm just surprised that all of these would have failed. 
___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 873 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UwEZkAy7b8E772FVQ:-P1UwEZkAy7b8E772FVR:b-r8lz6u) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L873)):*
> ```Python
> 
> def test_verse_range_matched_by_multiple_rows() -> None:
>     # A verse range picks up one text token per matched row, so all of them have to be read.
> ```

I'm unclear on what the point of this test is exactly. The combination of the test name and the comment aren't making sense to me. Is this really just a test that verse ranges are handled properly? Or is the point that there's duplicate text?
___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 804 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UuVFPD1GqUZAdwZmA:-P1UuVFQBoasztObuMCC:b8mtfiz) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L804)):*
> ```Python
> 
> def test_marker_behavior_disagrees_with_alignment_info() -> None:
>     # The behaviors in the alignment info are supplied by the caller and can disagree with the
> ```

Are both necessary? Having looked into this while making this fix, do recall if these is a clear reason we would need the behavior specified in both places?

Enkidu93 (commented):
Thank you for digging into this, Ben!

> @Enkidu93 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364#-:-P1UxGn195skN74B4kBp:b-eyg20e)* status: all files reviewed, 3 unresolved discussions (waiting on benjaminking and ddaspit).

benjaminking (commented):
> @benjaminking reviewed 1 file and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364)* status: all files reviewed, 3 unresolved discussions (waiting on ddaspit and Enkidu93).

___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 732 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UvlcOCQ4tLCJg9XyC:-P1V4_Fn9GKRrcHx-mGJ:b-p8yxat) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L732)):*

I'm assuming that all these tests were previously failing since you've added them or are covering a portion of uncovered code? I'm just surprised that all of these would have failed.
</blockquote></details>

I don't think the marker placement code was ever tested for unmatched markers.  I do believe all of these tests are covering behavior that was not tested previously (and I'm pretty sure these were all failing tests)
___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 804 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UuVFPD1GqUZAdwZmA:-P1VOnfE0cdMvRXx5crb:b-6p2vcy) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L804)):*

Are both necessary? Having looked into this while making this fix, do recall if these is a clear reason we would need the behavior specified in both places?
</blockquote></details>

I looked into it, and the behavior between the two is a little bit different, so we can't just use one or the other.  I do, however, think this design is unintuitive, but I'll try to address that in a separate PR.
___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 873 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UwEZkAy7b8E772FVQ:-P1V5Nma6I4FWhxh9Dqm:b-faq3fq) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L873)):*

I'm unclear on what the point of this test is exactly. The combination of the test name and the comment aren't making sense to me. Is this really just a test that verse ranges are handled properly? Or is the point that there's duplicate text?
</blockquote></details>

The marker placement algorithm was only pulling out the first text row when it was reconstructing a verse range.  It was essentially hard-coded for single verses.

Enkidu93 (changes requested):
> @Enkidu93 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364)* status: all files reviewed, 3 unresolved discussions (waiting on benjaminking and ddaspit).

___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 732 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UvlcOCQ4tLCJg9XyC:-P1WBN6fB0Q11EIvjexj:bmkerg) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L732)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

I don't think the marker placement code was ever tested for unmatched markers.  I do believe all of these tests are covering behavior that was not tested previously (and I'm pretty sure these were all failing tests)
</blockquote></details>

OK, great! Thanks for catching this. It's crazy how many tests there are and yet still things like this could be broken. 
___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 804 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UuVFPD1GqUZAdwZmA:-P1WBUG-6H8tWQooyWRo:b-tqtss1) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L804)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

I looked into it, and the behavior between the two is a little bit different, so we can't just use one or the other.  I do, however, think this design is unintuitive, but I'll try to address that in a separate PR.
</blockquote></details>

OK, feel free to just create an issue as well to track it as future works with your conclusions - I don't want to keep you from other things. 
___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 873 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UwEZkAy7b8E772FVQ:-P1WBaU747v4zSOb9Xz1:bkawwqd) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L873)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

The marker placement algorithm was only pulling out the first text row when it was reconstructing a verse range.  It was essentially hard-coded for single verses.
</blockquote></details>

I see, OK. I'm still not 100% following the comment or the test name. The 'verse range is matched by multiple rows'? The 'verse range picks up a text token per matched row'? Maybe I'm not reading them correctly but these don't make sense to me. If the important part is just that it's testing a verse range, we could probably just keep it at that. If it's important that the same tokens occur across parts of the range, then maybe mention that in the comment? Or if you feel that these are already clear and relevant, could you explain what they mean to me? :laughing:

benjaminking (commented):
> @benjaminking made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on ddaspit and Enkidu93).

___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 873 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UwEZkAy7b8E772FVQ:-P1WMwdO-qQdRadPs04G:b22n8m7) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L873)):*

I see, OK. I'm still not 100% following the comment or the test name. The 'verse range is matched by multiple rows'? The 'verse range picks up a text token per matched row'? Maybe I'm not reading them correctly but these don't make sense to me. If the important part is just that it's testing a verse range, we could probably just keep it at that. If it's important that the same tokens occur across parts of the range, then maybe mention that in the comment? Or if you feel that these are already clear and relevant, could you explain what they mean to me? :laughing:
</blockquote></details>

I have updated the name of the test as well as the comments.  It was definitely a bit of the Claude code idiolect.

Enkidu93 (approved):
:lgtm:

> @Enkidu93 reviewed all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine.py/364#-:-P1WUQ0J-YzFOObJrW8E:bnfp4nl)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on benjaminking and ddaspit).

___
*[`tests/corpora/test_place_markers_usfm_update_block_handler.py` line 873 at r1](https://reviewable.io/reviews/sillsdev/machine.py/364#-P1UwEZkAy7b8E772FVQ:-P1WUNWi3dznX4p_e9bW:b-f7sdrw) ([raw file](https://github.com/sillsdev/machine.py/blob/12e53e5e3eb73f85c5b4a7ff1b4985d2b33ccde0/tests/corpora/test_place_markers_usfm_update_block_handler.py#L873)):*
<details><summary><i>Previously, benjaminking (Ben King) wrote…</i></summary><blockquote>

I have updated the name of the test as well as the comments.  It was definitely a bit of the Claude code idiolect.
</blockquote></details>

Thank you - much better :grin:

